PR #13, opened 2024-10-07
Merge include_nvflare to dev-7-test-controller

oleschwen:
Using nvflare as a submodule is already part of dev-7-test-controller, so let's merge the branch completely rather than go via main.


Ultimate-Storm (approved):
Tested and works good, thanks!

PR #15, opened 2024-10-16
Dev 11 minimal application code

oleschwen:
This application code serves two purposes

- testing a training that runs quickly without depending on any data to be present
- example for writing own application code using pytorch.

PR #20, opened 2024-12-11
Get training via VPN to work

oleschwen:
Should be merged after merging #15 (builds on top of dev-11-minimal-application-code) and after resolving conflicts in application/jobs/minimal_training_pytorch_cnn/app/config/config_fed_server.conf and application/jobs/minimal_training_pytorch_cnn/app/custom/models/models_for_testing.py.

* extends documentation
* modifies timeouts
* extends startup kits,
* adds CNN of configurable size
* removes source directories of installed python packages from the docker images

Should be considered for merging together with https://github.com/KatherLab/NVFlare_MediSwarm/pull/1


_runTestsInsideDocker.sh
```diff
@@ -9,7 +9,7 @@
 # run unit tests of ODELIA swarm learning and report coverage
 export MPLCONFIGDIR=/tmp
 cd tests/unit_tests/controller
-python3 -m coverage run --source=/workspace/controller/controller -m unittest discover
+PYTHONPATH=/workspace/controller/controller python3 -m coverage run --source=/workspace/controller/controller -m unittest discover
```
Ultimate-Storm, line 12: This might still need to be discussed

Ultimate-Storm (approved):
With a rough look everything seems fine, thanks for the enormous efforts.

_runTestsInsideDocker.sh
oleschwen: Yes, that's not ideal yet.
For the training containers, I wanted to have just a single instance of the controller files at the installed location, so the respective Dockerfile deletes the installation directory.
For coverage testing, the Dockerfile copies the files into the image again to the location used here for the test.
I'd suggest reconsidering this in a future test refactoring and cleanup issue.

PR #24, opened 2025-02-11
dev-22 use same application code for local and swarm training

### LINKED ISSUE #22: Same 3dcnn_ptl code base for local and swarm training
A non-NVFlare version of the 3dcnn_ptl application code for local training would be useful to know that the local training uses essentially the same code as the swarm training.
Implementation could be similar to the `minimal_training_pytorch_cnn` application code.

AC:

- [x] 3dcnn_ptl application code capable of running both swarm and local training in the `nvflare-pt-dev:3dcnn` docker container
- [x] usage within and without startup kit documented

oleschwen:
See #22


oleschwen:
To test this without pushing images to the docker registry, 
* build the images but don't create and push the jefftud/… tags
* use the image without jefftud/ prefix when setting up the project via the dashboard
* use `--no-pull` for `docker.sh` from the startup kit

oleschwen:
Summarizing the changes:
* copy MediSwarm code into Docker image (to have one way of distribution for local training; also used for testing at that location)
* `main.py` in `3dcnn_ptl`application code can be run in swarm and local training mode (refactored in methods in new file)
* `docker.sh` in client startup scripts can run those two and the minimum example as an additional pre-flight check
* documentation in main README.md (maybe to be moved later)

In addition:
* fixed apt package versions and optimized Dockerfiles for `nfcore` and `3dcnn` containers

SteffenRenisch:
Tested at TUD; waiting for confirmation from UKA (test local training with respective option for docker script).

Ultimate-Storm (approved):
All tested, looks good to me

Issue #26, opened 2025-02-14
Versioning of Docker images


oleschwen:
We will likely use multiple versions of Docker images. To avoid confusion, we should use version numbers (probably more intuitive than, e.g., hash IDs) rather than a fixed name with varying images under that name.

AC

- [ ] implemented versioning of Docker image
- [ ] documented for developers how to upgrade versions

oleschwen:
Following #22, the image would contain the code for local training. To ensure consistency between the code for swarm traning and local training, we may want to to avoid accidentally using non-matching versions.
* We could check, before starting the admin console, that the code version to be sent out by NVFlare matches the version contained in the 3dcnn image for local training. See afc96c299d0a8b42899458167bda8988a2a0e3c2.
* If we start the admin console from the nfcore docker image (see #27), we could use a fixed version of the application code copied into the docker image. See 9b1ad8ace504aea3762ed64c5816da26196a99e5.
* We could also consider using a common docker image rather than nfcore used for the admin and 3dcnn for the sites. (Con: image contains more than needed for the respective purpose, Pro: only one image to maintain). See 9da4fe1f5a339528a6761dfec16c8807bc7dc425
* We could also run the tests in the single, combined docker image. See, e.g., d79e91ed9ee81c6945a12f50d135922f249dd588.
All variants are technically feasible and work in a MEVIS test setup. Some cleanup will be necessary after reverting to each of these commits and the READMEs will need to be updated accordingly.

oleschwen:
In any case, we should have one script that ensures a consistent version number across everything to be built (i.e., docker image(s), startup kits, anything else?).

SteffenRenisch:
Separate the questions:
- do we want to combine the docker images for the admin console and for the clients into one image? - Yes
- do we want to distribute the application code from the image? - Yes

additionally: 
- do we want to check the application code before execution against the version in the (client) docker image

oleschwen:
The package name we use for our NVFlare fork, nvflare_nightly-2.3.0.dev<current date>, looks like an "official" package rather than our ODELIA fork, which is confusing. There are no obviously good package and version names. It is forked from nvflare 2.4.1, not 2.3.0, but it's not our version 2.something and it's not vanilla nvflare 2.something. It is also somewhat independent of our docker versioning. Also, the current date is informative for our purposes, but not necessarily a unique identifier.
I'd suggest `nvflare_mediswarm` version `1.0.dev<current date>`.

[...]

Issue #32, opened 2025-03-03
Setup swarm training on ODELIA data


SteffenRenisch:
Make sure every site is able to use their own data using the docker image provided.

We need to provide startup kits, initiate training.

Other minor changes we noticed along the way when testing the DUKE training in a bigger swarm:

- [x] `buildStartupKits.sh` should fail in case of local changes
- [x] remove unused (unmaintained and untested) scripts from startup kits (aws, azure, start, sub_start)
- [x] only ask for required options in `docker.sh`, document accordingly in `README.md`

oleschwen:
Can we make it easier for particpants to use the correct (and not an outdated) startup kit, e.g., by making "obtain startup kit" and "start client using the obtained startup kit" a one-step procedure?
The client will fail (due to certificates) if it does not match the server version. However, this is the second (or third) step for the participants, we want to run local preflight checks before attempting to connect to the swarm.

oleschwen:
The docker-in-docker proof of concept mode does not work so far, see #39. If we don't get it to work, we can possibly make the Docker image smaller by not installing Docker and dependencies in the first place.

oleschwen:
Requiring root access for creating the scratch directory and changing ownership might not be available when starting a client. We could, instead, require the directory to exist with suitable ownership and permissions. However, this would add additional steps to what should ideally be a one-step procedure.

oleschwen:
`envsetup_scripts/setup_vpntunnel.sh` does not give the intended error message if the site is unspecified. Will fix it.
Actually, it would be helpful to provide vpn connection scripts without the need for cloning the entire repository.

oleschwen:
The image contains conda configured to use a channel users may not have a license to use. Conda is not used, though, so we should uninstall it in the image.

PR #38, opened 2025-03-11
Dev 14 document prototype setup

### LINKED ISSUE #14: Document Setup for Running Swarm Training
Trying to get a two-site swarm training on the Duke dataset to run, I noticed the following:

The jefftud/nvflare-pt-dev:3dcnn (a3bd4e8a9577) image does not work for me, I think there are two issues: a psutil version problem as Steffen described in July, and the PYTHONPATH variable that should be set to /workspace/controller/controller. I've continued with a locally built image from docker_config/Dockerfile_3dcnn from the dev-7-test-controller branch (planned to be merged to main, #7). The image should be updated accordingly.

Moreover, simulation mode does not work out of the box, but requires having a dataset available and split somehow (or just symlinked to the full folder) matching the simulated node names. This should be explained and/or included in the scripts.

This might resolve part of #2.

AC
- [ ] following README.md and using scripts allow running simulation and POC modes successfully.

oleschwen:
See #14. 

Are there changes in this branch we want to preserve and merge to main?

* preparation of DUKE dataset (that may become partially obsolete with the ODELIA data format, may serve as a template for using local data in local testing of swarm code, but would need to be adapted for this purpose)
* notes on dashboard usage → obsolete, we want to use command-line provisioning
* VPN usage documentation → obsolete, instructions added in #26
* remove unused, untested and unmaintained azure scripts from startup kits → would be useful to keep


oleschwen:
Withdrawn – startup kits will be cleaned up further as part of #32

PR #47, opened 2025-03-26
Dev 32 setup swarm training on odelia data

oleschwen:
Minor changes we noticed along the way when testing the DUKE training in a bigger swarm, see #32

* improved script to build docker image and startup kits
* improved `docker.sh` in startup kits
* removed unused cloud start scripts
* also print output for clients
* improved VPN setup script
* uninstall conda in image
* updated version numbers of apt packages
* corresponding updates in README

PR #48, opened 2025-04-23
fix missing version numbers for last-in-line apt packages

oleschwen:
I missed some version numbers of apt packages in #47. This PR should fix these.


Ultimate-Storm (approved):
approved

PR #49, opened 2025-04-24
further corrections of apt package versions

oleschwen:
The previous version number update in 8d7c1c098135ae76b0a2532a5728611c9566d067 unfortunately still did not work due to some incorrect version numbers. This one should work.


Ultimate-Storm (approved):
Works for me, good to go

PR #54, opened 2025-05-19
53 do not git directory in image

### LINKED ISSUE #53: Do not .git directory in image
We should not include local branches and unversioned files in the Docker images.

oleschwen:
see #53

Should be reviewed and merged after #49


Ultimate-Storm (approved):
Good to go

PR #57, opened 2025-05-23
Apt package version update

oleschwen:
We need another update of an apt package version


Ultimate-Storm (approved):
Tested and works for me!

PR #59, opened 2025-05-28
update apt package versions

oleschwen:
We need another update of apt package versions for building the Docker image.
Also, no longer use docker cache by default.


Ultimate-Storm:
The building is failing

oleschwen:
Should work again.

Ultimate-Storm (approved):
Works for me!

PR #64, opened 2025-06-12
Update apt package versions 20250612

oleschwen:
- Some apt packages need to be installed as newer version number, updated those. 
- Removed two packages already installed at latest version in base image.


oleschwen:
Sorry, based on wrong branch.

PR #88, opened 2025-07-14
Omit confusing number of labels

oleschwen:
- Omit confusing number of potential labels (not number of different labels present in dataset)
- Updated apt package versions for successful build
- Use http also in test setup (as we have done in the production setup for a while)


Ultimate-Storm (approved):
approved

Issue #85, opened 2025-07-11
Implement Feedback on README files


oleschwen:
The README files for the different roles contain todos and should generally be improved based on the feedback from the first ODELIA-wide swarm training using MediSwarm.

oleschwen:
Own findings

- [x] avoid asking for sudo password when starting client
- [x] output version number in log

oleschwen:
continued in #102

PR #96, opened 2025-07-21
Improvements Based on 1.0 Test Feedback

oleschwen:
First changes based on own and sites' feedback received from testing dummy training, preflight check, and local training:

- pre-trained ResNet variants can be used as alternatives to MST
- log which code version was used
- improved README for participants
- streamlined build of Docker image and startup kits: omit copy&delete of potentially large workspace folder
- avoid need for admin password when starting client (can only write to writeable location)


oleschwen:
This PR partly addresses #85 but does not resolve the issue yet. We should create a new branch for further feedback and improvements after making the current updated README visible.

Ultimate-Storm (approved):
Approved, thanks for the efforts

PR #110, opened 2025-08-21
removed validadion of pinned versions

### LINKED ISSUE #109: update-apt-versions workflow should not remove version numbers
actual behavior: in #108, version numbers of apt packes were removed
expected behavior: version numbers should be updated to to the currently available version

This could be due to the check for invalid versions in scripts/ci/update_apt_versions.sh using a different/outdated cache of available version than what is used in the docker build.

oleschwen:
This PR is intended to solve the problem of disappearing version numbers of pinned apt packages.


Ultimate-Storm (approved):
approved

PR #113, opened 2025-08-26
List packages in Docker image and their licenses

### LINKED ISSUE #83: List licenses
Currently, the server startup kit has a mechanism to list licenses of pip packages. This should be extended:

- [x] all startup kits allow listing licenses of installed packages
- [x] listing of licenses includes apt packages, pre-trained weights
- [x] verified that this is all third-party content in the images

oleschwen:
* all startup kits provide listing of packages and their licenses
* dropped Docker and dependencies from container, they are not needed

This introduces the option `--list_licenses` to `docker.sh` in all startup kits to list all apt and pip packages contained in the image with their respective licenses, as well as the license of the contained pre-trained weights. Moreover, we do not need Docker installed in the container, thus dropped Docker and its dependencies.


oleschwen:
should be merged after #118 and resolving likely conflicts

Ultimate-Storm (approved):
tests passed, looks good to me

PR #118, opened 2025-09-09
extend integration tests

### LINKED ISSUE #104: Testing Startup Kits
We currently have unit tests and and simulation/proof-of-concept mode tests. We should also test creating and using the startup kits (and possibly further things worth testing along the way).

oleschwen:
This PR extends the integration tests by several aspects, in particular by starting a swarm with server and client nodes on a single machine.

Additional changes:
* Containers now have more specific names
* Slightly updated documentation


oleschwen:
CI pipeline took about 9 minutes (https://github.com/KatherLab/MediSwarm/actions/runs/17798477520/job/50591721053?pr=118) with an unusally slow Docker build, so it should typically be a bit faster.
Two steps are not included
* pushing to and pulling from a local test container registry (4 minutes with limited value for testing: we will notice if a productive pull/push fails and that would likely be due to reasons not present in the test)
* starting a swarm and submitting a job via the admin console (requires `expect` and potentially further adaptation of the test environment; this would be desirable, possibly outside this PR).

.github/workflows/pr-test.yaml
```diff
@@ -50,37 +47,52 @@ jobs:
...
+
+      - name: Run dummy training in simulation mode
+        continue-on-error: false
         run: |
-          ./runTestsInDocker.sh prepare_dummy_trainings
-          echo "Dummy training project prepared"
+          ./runIntegrationTests.sh run_dummy_training_simulation_mode
 
-      - name: Run dummy training
+      - name: Run dummy training in proof-of-concept mode
```
Ultimate-Storm, line 70: I wonder why do we need to test the POC mode here?

Ultimate-Storm (approved):
Thanks for your work! 
The PR CI looks good and good to go

.github/workflows/pr-test.yaml
oleschwen: This is mainly a test whether we have a working documentation by example/blueprint how the proof-of-concept mode can be used for ensuring that other application code is swarm-ready.
It indeed does not say a lot about the minimal example, which is also run stand-alone, in simulation mode, and in a proper swarm training in separate tests.

PR #126, opened 2025-09-30
POSTPONED: VPN from within container

### LINKED ISSUE #122: Test VPN connection from within container
One partner has issues with the setup of the VPN connection on the host. This feature shall test whether setup of the VPN within the container (instead of on the host) solves the problem.

First test: setup of container with VPN connection setup within container, test with two container e.g. at TUD and Bremen.

oleschwen:
Test setup for connecting to VPN from inside the MediSwarm containers rather than on the host, including a setup for locally running an openvpn server. Postponed, as we do not need this for now.
The branch contains credentials used for the test setup, warnings about this can be ignored.


oleschwen:
Corresponding issue #122 was closed.

PR #138, opened 2025-10-17
Fix usage of GPU when running all tests

oleschwen:
This PR 
* fixes the problem that ./runIntegrationTests.sh did not allow running all tests due to inconsistent use of the GPU_FOR_TESTING environment variable
* extends the documentation of the integration tests
* includes the NVFlare unit tests in running "all" tests
* includes the swarm tests in running "all" tests


Ultimate-Storm (approved):
LGTM

PR #146, opened 2025-10-23
Extend integration tests

### LINKED ISSUE #144: Extend Integration Tests
There are a few more requirements (following #104) we should test as part of the integration test. This issue and the related branch collect extensions of the test that are currently not critical, but nice to have.

oleschwen:
Additional tests:
* minimal example actually uses GPU
* preflight checks take less than one minute
* more expected output in server and client logs
* admin console with invalid certificate cannot connect
* startup kit has option for local training

Further changes:
* kill orphaned containers at the end of the test if steps in the workflow have failed and containers have not been killed as part of the respective step


Ultimate-Storm (approved):
looks great

PR #159, opened 2025-11-17
147 check if custom swarm controllers are needed

### LINKED ISSUE #147: Check if custom swarm controllers are needed
To reduce future documentation, maintenance, and test efforts, it would be helpful to use the NVFlare swarm client and server controller classes instead of our custom ones, which mainly implement additional exception logging. We should investigate which changes are really necessary and how/where to best implement them.

AC:

- [x] decided if any changes to NVFlare controller classes need to be kept
- [x] verified that training with NVFlare controller classes works in test setup
- [ ] verified that training with NVFlare controller classes works in ODELIA setup

oleschwen:
* replaced usage of custom swarm client and server controllers by NVFlare controllers
* removed custom controllers
* removed corresponding unit tests
* added local openvpn container setup for manual testing


Ultimate-Storm (approved):
All checks passed with the removal of codes. Addition of new codes shall work fine. Approved!

PR #163, opened 2025-11-27
Make cleanup script for old images compatible with Docker 29.0

### LINKED ISSUE #162: Make image cleanup script compatible with Docker 29.0
The dev util script to delete old versions of ODELIA Docker images no longer works for Docker 29.0 due to a different output of `docker image list`.

AC

- [x] script works with Docker 29.0
- [ ] to be decided: do we need backwards compatibility?

oleschwen:


Ultimate-Storm (approved):
approved

PR #181, opened 2026-01-16
Store ground truth and class probability values in csv when validating

### LINKED ISSUE #177: Store GT and class probability values per image during validation
Currently the AUC is computed and logged for evaluation. We want to store per image the GT and the class probabilities per label in order to compute additional metrics retrospectively.

oleschwen:
Output ground truth values and class probabilities for training and validation images in the ODELIA ternary classification job.
This data is written to four separate csv files for the distributed aggregated model at the beginning of the round (swarm training only) and for the site-specific model after each epoch (swarm and local training).
We currently have no integration test where creation of these files can be checked with reasonable effort. The job is currently only run in simulation mode, after which we have no access to the respective temporary output directory.

### DIFF
--- a/application/provision/project_MEVIS_test.yml
+++ b/application/provision/project_MEVIS_test.yml
@@ -4,7 +4,7 @@ description: >
   Test setup.
 
 participants:
-  - name: odelia-vm-a-localvpn
+  - name: localhost
     type: server
     org: MEVIS_Test
     fed_learn_port: 8022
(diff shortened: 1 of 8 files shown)


application/provision/project_MEVIS_test.yml
oleschwen, line 1: This was needed for local testing but should be irrelevant for the production swarm.

Ultimate-Storm (commented):
approved

Ultimate-Storm (approved):

PR #207, opened 2026-02-26
206 optional output for dataset debugging

### LINKED ISSUE #206: Optional output for dataset debugging
* We should output more details on the datasets loaded so that clients can debug data usage. 
* To avoid leaking information about private data, this should not be logged by default, but only when specifically requesting this.
* One natural location for such output is the data access preflight check.
* We may want to output discrepancies between images available and metadata tables (which may or may not be intended)

oleschwen:
* Added plausibility checks for the datasets, logging errors/warnings for discrepancies in UIDs or between image data and metadata
* An optional argument for the preflight check also logs the individual UIDs (might leak information about dataset if logs are shared, thus not default behavior)


Ultimate-Storm (approved):
Looks good to me

PR #283, opened 2026-04-10
fix/apt version update main

Ultimate-Storm:
- **fix: handle apt update before pinned installs**
- **fix: allow startup kit builds without live sync secrets**
- **fix: repin apt packages after rebuild**
- **chore: refresh apt pins in Dockerfile_ODELIA**


oleschwen (approved):
looks good

PR #288, opened 2026-04-13
228 Allow running ODELIA containers in parallel

oleschwen:
This PR introduces the option `--container_name` for the client startup kit `docker.sh` for specifying a custom name for the Docker container. The default name based on the startup kit version remains unchanged, but we can now run multiple containers for the same startup kit in parallel, e.g., for running multiple local trainings on the same machine.


oleschwen:
I would not include this test in the CI, it adds too little value for an additional minute of CI run time.

Ultimate-Storm (approved):
Approved

PR #297, opened 2026-04-16
289 Consistent caching of pre-trained model weights

### LINKED ISSUE #289: ODELIA: improve handling of pre-trained models
We currently have two ways of handling pre-trained models when building the ODELIA Docker image:
* cache in directory created inside local repository copy
* download during build to location in /tmp

We should not clutter the repository folder with such a cache and we should not rely on /tmp having enough storage space for a temporary download. Ideally, we should not rely on external resources being available when building images. Instead, we use consistent caching with a configurable location and the option to make caching optional.

oleschwen:
This PR makes the way of caching pre-trained model weights when building Docker images consistent for all models. If an environment variable MEDISWARM_BUILD_CACHE_DIR is set, that location is used as a persistent cache, otherwise they are downloaded to a tempoary directory and deleted after the build is complete. This no longer creates a cache inside the local repository clone for the dino pre-trained weights.


Ultimate-Storm (approved):
Approved

PR #304, opened 2026-04-21
Revert "Improve ODELIA dataset plausibility checking"

oleschwen:
Reverts KatherLab/MediSwarm#298
This does not work as intended yet, the check reports false positive discrepancies for the public challenge data.


Ultimate-Storm (approved):
Approved, thanks!

PR #311, opened 2026-04-29
310 Extend integration test

### LINKED ISSUE #310: Extend integration test
As part of #234 (which is not to be merged in main), extensions of the MediSwarm integration are being implemented. These extensions would be beneificial in main.

oleschwen:
* check that Docker/GPU preflight check fails if no GPU available
* added possibility to run NVFlare integration test
* temporarily disabled checking for publicly available files on github


Ultimate-Storm (approved):
Approved

PR #327, opened 2026-05-13
326 ODELIA swarm training of challenge jobs

### LINKED ISSUE #326: ODELIA: swarm training of challenge jobs
Training the swarm models (using the public challenge data) as swarm training jobs currently does not work in a local setup.

AC:

- [ ] All five models trained successfully in local training in test setup
- [ ] All five models trained successfully in swarm training in test setup
- [ ] Shared results as baseline for distributed training

oleschwen:
This PR
- allows submitting all challenge models and the MST as swarm training jobs
- speeds up computation of predictions per image by using CUDA
- increases a timeout to prevent failing trainings due to different nodes taking different wall clock time per round
- uses `--shm-size=16g` for swarm clients (like we already do in the integration tests) to avoid memory issues


oleschwen:
Might create conflichts with #339

Ultimate-Storm (approved):
Good to merge, although I will need to increase the server timeout in another PR based on experiments findings.

PR #340, opened 2026-06-10
Guard ODELIA normalization against zero-std images

Ultimate-Storm:
## Summary

Add a cleaned robustness patch for the collaborator-reported ODELIA local-training crash around masked Z-normalization:

- `ZNormalization` now returns a zero tensor instead of raising when the mask has fewer than 2 voxels or when TorchIO normalization reports zero standard deviation.
- Percentile tensors are created on the same device/dtype path as the masked image tensor.
- A local JSONL diagnostic is written to `$SCRATCH_DIR/odelia_zero_std_guard.jsonl` or `/scratch/odelia_zero_std_guard.jsonl` with the detailed image path and guard reason.
- Stdout only emits a short sanitized marker by default, so full image paths/UIDs are not broadcast through live monitor logs. Full stdout details remain opt-in via `ODELIA_ZERO_STD_GUARD_LOG_DETAILS=1`.
- `ODELIA_Dataset3D` restores the source `path_img` on tensor-backed cached images before transforms, so local diagnostics can still identify the problematic source file.

## Why

A collaborator had to patch local files to avoid crashes during local training. We should ship the fix through the normal PR/release/startup-kit flow rather than asking partners to overwrite files inside containers or startup kits.

## Validation

- `python3 -m py_compile application/jobs/_shared/custom/data/augmentation/augmentations_3d.py application/jobs/_shared/custom/data/datasets/dataset_3d_odelia.py`
- `.venv/bin/pytest -q tests/unit_tests/test_odelia_zero_std_guard.py tests/unit_tests/test_data_module.py tests/unit_tests/test_threedcnn_ptl.py` -> 28 passed
- `git diff --check`

## Follow-up after merge

- Rebuild all-sites startup kits from `main`.
- Push the new Docker image.
- Ask the collaborator to rerun the same local-training command that previously crashed and confirm whether `odelia_zero_std_guard.jsonl` is produced only locally when problematic images are encountered.

### DIFF
--- a/application/jobs/_shared/custom/data/augmentation/augmentations_3d.py
+++ b/application/jobs/_shared/custom/data/augmentation/augmentations_3d.py
@@ -1,3 +1,7 @@
+import json
+import os
+import time
+from pathlib import Path
 from typing import Union, Optional, Sequence
 
 import torchio as tio
@@ -17,6 +21,9 @@ import random
 
 import torch
 import torch.nn.functional as F
+
+
+_TRUTHY_ENV_VALUES = {"1", "true", "yes", "on"}
 
 
 class Random3DAugmentation:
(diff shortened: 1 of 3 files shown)


Ultimate-Storm:
Superseded by #368.

@oschwen's review here was the right call: silently replacing a degenerate image with a **zero volume** (and training on it) hides the problem rather than fixing it. #368 takes the approach discussed:

- **Skip + resample** a degenerate/corrupt input instead of zero-filling (a typed `DegenerateImageError`; `__getitem__` omits the bad sample and draws a different one — never a zero image), and **aborts if >5% of inputs are bad** so a systematic problem is surfaced.
- A **preflight input-check** (`find_degenerate_inputs`) that lists constant **and** corrupt inputs to report to sites / exclude via `$ODELIA_EXCLUDE_UIDS_FILE`.
- The privacy-safe local logging idea from this PR is kept (`odelia_bad_inputs.jsonl`, UID off the console).

Also: a 3-site scan via the production loader (RSH/MHA/USZ, ~7,100 train/val/test volumes) found the models' actual inputs **all clean** — so the zero-std cases are bad *inputs* (constant/corrupt), not crop artifacts, and not a widespread issue at the accessible sites. The remaining UMCU case can be settled by running `find_degenerate_inputs` on UMCU's data.

Closing in favor of #368.

PR #350, opened 2026-06-22
Let CI fail if errors occurred

### LINKED ISSUE #349: CI should fail in case of errors
CI currently passes even though part of the tests fails, e.g., https://github.com/KatherLab/MediSwarm/actions/runs/27896067032/job/82547686021, "Run 3DCNN training in simulation mode".
Such failure should be visible as failed CI.
(Note: we need to distinguish between expected warnings and errors required to pass tests and actual failures.)

oleschwen:
This PR makes the CI fails if errors occurred in trainings in simulations and proof-of-concept.
Follow-up issues: #351 and #352.

### DIFF
--- a/scripts/ci/runIntegrationTests.sh
+++ b/scripts/ci/runIntegrationTests.sh
@@ -489,8 +489,8 @@ run_3dcnn_simulation_mode () {
 
     echo "$OUTPUT"
 
-    echo "TODO adapt criterion for passing"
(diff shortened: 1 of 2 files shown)


oleschwen:
https://github.com/KatherLab/MediSwarm/actions/runs/27951294683/job/82708575485 demonstrates that error now leads to failed CI

Ultimate-Storm (approved):

Issue #352, opened 2026-06-23
ODELIA Integration test should test current default (ideally all) productive models


oleschwen:
Following #349:
The integration tests currently test the MST. Instead, they should use at least the current default model for local and swarm training, ideally all implemented models.
We should consider which tests should be part of the CI to avoid excessively long feedback cycles, e.g., training the current model in CI and all implemented models in an extended test we could schedule, e.g., once per week.

- [x] implement preflight checks and example trainings for all available models
- [x] set up workflow for periodically running "extended" integration test
- [x] check if any parts of integration tests currently not part of CI should be run periodically
- [x] check if any parts of integration tests currently part of CI should only be run periodically

oleschwen:
Also check whether pre-trained weights can be used offline (not the case for ResNet18 at the moment), see #376

oleschwen:
resolved via #561 and #547: weekly test now contains preflight checks and swarm trainings for most of the available models

PR #354, opened 2026-06-24
CI: 3DCNN simulation test silently passes on an aborted run (#353)

### LINKED ISSUE #353: CI: 3DCNN simulation test silently passes on an aborted run (no success assertion + empty MODEL_NAME)
## Problem
The `run_3dcnn_simulation_mode` CI step prints fatal errors but **CI still passes** — a false pass. Surfaced while reviewing PR #305's CI run.

In `tests/integration_tests/_run_3dcnn_simulation_mode.sh` the federated simulation **aborts**:
```
ERROR - Aborting current RUN due to FATAL_SYSTEM_ERROR received:
        failed to configure clients ['client_A','client_B']: only 0/2 configured, need 2
ValueError: failed to instantiate class: ValueError: Unsupported model name: .
```
Two compounding bugs:
1. **Empty `MODEL_NAME`** — the persistor's `models.models_config.create_model` is called with an empty model name (`Unsupported model name: .`). CI passes `-e MODEL_NAME=${MODEL_NAME:-}` (empty when unset) and `pr-test.yaml` doesn't set it for this step, so both clients fail to configure and the run aborts.
2. **No success assertion** — the test just runs `nvflare simulator …` and relies on its exit code, but `nvflare simulator` returns **0 even when the run aborts with `FATAL_SYSTEM_ERROR`**. With `set -e` never tripping, the step exits 0 and CI goes green.

Net effect: a swarm simulation that did **nothing** (0/2 clients configured) is reported as a pass — masking real regressions.

## Proposed fix
1. Set a concrete `MODEL_NAME` for the 3DCNN sim step (a ternary/challenge model), **or** have the ternary persistor fall back to a valid default when `MODEL_NAME` is empty.
2. Make the test **assert real success** — grep the simulator output for a completion marker and/or fail on `FATAL_SYSTEM_ERROR` / `0/N configured` — so an aborted run fails CI instead of silently passing.

## Files
- `scripts/ci/runIntegrationTests.sh` (`run_3dcnn_simulation_mode`, `-e MODEL_NAME=${MODEL_NAME:-}`)
- `tests/integration_tests/_run_3dcnn_simulation_mode.sh`
- `.github/workflows/pr-test.yaml`

Ultimate-Storm:
Fixes #353.

**Problem:** the `run_3dcnn_simulation_mode` CI step printed fatal errors but passed:
1. `MODEL_NAME` was empty → `create_model()` got `Unsupported model name: .` → both clients failed to configure → run aborted (`FATAL_SYSTEM_ERROR ... only 0/2 configured`).
2. `nvflare simulator` exits **0 even on a FATAL abort**, and the test had **no success assertion**, so `set -e` never tripped → false pass.

**Fix** (`tests/integration_tests/_run_3dcnn_simulation_mode.sh`):
- Default `MODEL_NAME` to `ResNet18` (a valid ternary 3D-CNN) when unset/empty.
- Assert real success: **fail** on `FATAL_SYSTEM_ERROR` / `failed to configure clients`, on a non-zero simulator exit, or if no `FL_global_model.pt` was produced; print `=== 3DCNN Simulation Mode PASSED ===` only on real success.

### DIFF
--- a/.github/workflows/pr-test.yaml
+++ b/.github/workflows/pr-test.yaml
@@ -11,14 +11,20 @@ on:
 permissions:
   contents: read
 
-# Only one validation run at a time on the self-hosted runner to prevent
-# resource contention (GPU, Docker, disk) that causes flaky test failures.
+# Cancel superseded validation runs for the same PR/ref. The GPU-heavy
+# validate-swarm job below also queues globally to avoid self-hosted GPU
+# contention across different PRs.
 concurrency:
   group: pr-validation-${{ github.event.pull_request.number || github.ref }}
   cancel-in-progress: true
 
 jobs:
   validate-swarm:
+    # The self-hosted runner has one shared GPU; running multiple validation jobs
+    # at once can OOM otherwise independent PRs.
+    concurrency:
+      group: mediswarm-self-hosted-gpu-validation
+      cancel-in-progress: false
     runs-on: [self-hosted, ci]
     timeout-minutes: 60
 
--- a/scripts/ci/runIntegrationTests.sh
+++ b/scripts/ci/runIntegrationTests.sh
@@ -486,15 +486,18 @@ run_3dcnn_simulation_mode () {
     # requires having built a startup kit and synthetic dataset
     echo "[Run] Simulation mode of 3DCNN training in Docker (capturing output)"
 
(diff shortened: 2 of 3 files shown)


oleschwen (approved):
Looks good.
Default model is changed to ResNet18, not one of the productive models, but we have #352 to re-consider which model(s) to use.

oleschwen (approved):
conflict resolution looks good

Issue #348, opened 2026-06-19
Swarm: GoodAccess VPN instability drops tunnels at multiple sites and aborts runs


Ultimate-Storm:
## Problem
The GoodAccess VPN (gateway `14357-de-9957-tt8rw.accessgate.cloud`) intermittently drops site tunnels. A drop **during the swarm's peer-to-peer round gather** returns `ERROR` → fatal abort of the whole run. Now confirmed at **two different sites**, which points to a **gateway-side** problem rather than a single node.

## Evidence (June 2026)
- **UKA** (`titan`, NetworkManager-OpenVPN, this gateway): recurring OpenVPN `--ping-restart` drops. On Jun 17 the client exited fatally on a failed `--up/--down` helper and **stayed down ~2.5 h** with no auto-reconnect (the VPN password was agent-owned, so only an interactive session could re-supply it). Same trigger seen Jun 13.
- **VHIO** (`openvpn --config .../accessgate.cloud.ovpn`, same gateway): tunnel dropped ~04:46 UTC Jun 19 and re-handshook at 04:52 — a ~6-min drop that aborted run `696648fe` at round 8 during the gather.
- GoodAccess support engaged; they reported the UKA outage was node/path-side and requested client logs (provided).

## Mitigation already applied (UKA)
- VPN secret moved to **system storage** (`password-flags=0`) so a background process can reconnect.
- Added a **systemd watchdog** (`goodaccess-vpn-watchdog.timer`, 30 s) that re-ups `tun0` only when it has dropped → ~30 s recovery vs ~2.5 h. Non-disruptive (no-op while healthy).

## Proposed fixes
1. **Standardize a resilient, unattended VPN client across all sites**: system-stored credentials, auto-reconnect (`keepalive` / `ping-restart`, `Restart=always` or the watchdog pattern), documented as a client-setup step. VHIO's plain-`openvpn` daemon already self-recovers (~6 min); UKA needed the watchdog — converge them.
2. **Escalate gateway stability with GoodAccess** — multi-site evidence supports a gateway-side fix; explore keepalive / `explicit-exit-notify` tuning or a dedicated gateway.
3. Pair with the **fault-tolerance** issue (retry/tolerate a brief peer drop in the gather) and **warm-continue** (bound the cost of any drop) so a short blip can no longer destroy a multi-hour run.
4. Add a VPN-robustness checklist to the participant guide.

## References
[...]

PR #345, opened 2026-06-18
Harden ODELIA swarm: 24h timeouts + wait-for-all, preflight checks, cgroupfs fix, docs

Ultimate-Storm:
## Why
Consolidates the fixes from a real 8-site ODELIA swarm campaign so future builds/runs are robust and participants can self-detect/fix problems **before** a run. **Supersedes #339** (24 h timeouts); builds on **#327** (merged — per-model swarm configs, `--shm-size`, MODEL_NAME gating). Full catalogue: `docs/SWARM_FAILURE_MODES.md`.

## What

**Config — all 6 `challenge_*` jobs + `ODELIA_ternary_classification`**
- server (`config_fed_server.conf`): `max_status_report_interval` 300 s → **24 h**, `progress_timeout` 8 h → **24 h**, `start_task_timeout`→60 min, `configure_task_timeout`→30 min. (A slow site is silent for its whole round; the 5-min status interval caused the `didn't report status` abort — F1.)
- client (`config_fed_client.conf`): `learn_task_timeout` 8 h → **24 h**; **`min_responses_required` 2 → 6 (wait-for-all)** so a slow site is never left behind (prevents `MODEL_UNRECOGNIZED` desync — F5). Documented tradeoff: a mid-round drop then stalls until `learn_task_timeout`; set this to the participating-client count.

**Preflight checks** — `docker_config/master_template.yml` (`_preflight_host_checks`), run for `--dummy_training`/`--preflight_check`/`--start_client`, print a PASS/WARN/FAIL block with remediation for: GPU usable in container (F2), **systemd cgroup-driver risk** (F2), read-only cache path (F4), server reachability (F6), and a stale `daemon_pid.fl` (F3, auto-cleared when no client is running).

**`scripts/client_node_setup/fix_docker_cgroupfs.sh`** — durable host fix for the NVML "Unknown Error" GPU drop (daily `daemon-reload` on cgroup v2 + systemd driver): switches Docker to `cgroupfs`, auto-reverts if Docker won't restart, validates a `daemon-reload` no longer strips the GPU.

**Docs** — new `docs/SWARM_FAILURE_MODES.md` (F1–F7 + operator playbook); `docs/TIMEOUTS.md` updated (24 h values, `max_status_report_interval`/`min_responses_required`, slow-node & sync section); participant guide (pre-run checklist + 4 troubleshooting entries); developer guide (operating & hardening section).

## Notes
[...]

PR #355, opened 2026-06-24
VPN resilience: faster tunnel-down detection + mandate auto-recovering service (#348)

Ultimate-Storm:
Refs #348.

Long swarm runs kept aborting on a single site's VPN drop. The repo already has the right primitives — `mediswarm-vpn.service` (`Restart=always`) and `vpn_health_monitor.sh` — but the failing sites used ad-hoc GUI / bare `openvpn --config` setups, and the monitor only reacted after 3 failed gateway pings (~3 min).

**Changes:**
- `vpn_health_monitor.sh`: **immediate restart when the tunnel interface (`tun0`) is down / IP-less** (the definitive drop signal), in addition to the existing gateway-ping threshold; timer tightened to **30 s**; new `VPN_IFACE` env (default `tun0`).
- `README.participant.md`: make the **auto-recovering systemd setup the recommended path** (`setup_vpntunnel.sh -d <site> -n -s` + `vpn_health_monitor.sh --install-timer`) and add a *VPN drops mid-run / 'deemed disconnected'* troubleshooting entry.

Together with the 24 h round timeouts (#345), a tunnel restored in ~30 s lets a run ride out a transient blip instead of aborting. (GoodAccess gateway-side escalation is tracked separately / external.)

### DIFF
--- a/assets/readme/README.participant.md
+++ b/assets/readme/README.participant.md
@@ -158,6 +158,15 @@ To have a baseline for swarm training, train the same model in a comparable way 
 
 1. Connect to VPN as described in [VPN setup guide(GUI).pdf](../VPN%20setup%20guide%28GUI%29.pdf) (GUI) or [VPN setup guide(CLI).md](../VPN%20setup%20guide%28CLI%29.md) (command line).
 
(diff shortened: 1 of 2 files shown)


oleschwen (approved):
looks good

Issue #347, opened 2026-06-19
Swarm: warm-continue / periodic checkpointing to recover an aborted run without losing progress


Ultimate-Storm:
## Problem
When a swarm run aborts (see the fault-tolerance issue), **all progress is lost** — there is no resume, and re-submitting restarts from random initialization. With 15–17 h runs repeatedly dying at round 7–10, this has prevented any full 20-round run from completing.

Worse, the global model is **only persisted on the rotating aggregator** each round, so the latest global can be **stranded on the node that crashed**: after `40c0948f` aborted at round 10, the round-10 global was on the crashed UMCU_1 (unreachable); the latest *recoverable* global was ~round 8 on MHA_1.

## Proposed fixes
1. **Documented + tooled warm-start.** The client persistor is `PTFileModelPersistor`, which accepts `source_ckpt_file_full_name`. Provide a helper that (a) fetches the latest `best_FL_global_model.pt`, (b) bundles it into the job, (c) sets `source_ckpt_file_full_name`, (d) reduces `num_rounds` to the remainder — i.e. resume from the last good global.
   - Note: `create_model(pretrained_path=...)` only loads the **UNet encoder**, so it is *not* a full-model warm-start — use the persistor's `source_ckpt`.
2. **Shorter jobs + chained warm-starts.** Submit e.g. 5-round jobs and chain them; a failure then costs ≤5 rounds instead of the entire run.
3. **Central / redundant global persistence.** Persist the aggregated global each round to a location independent of which node crashes (server-side copy, shared mount, or all clients), so the latest is never stranded on a failed node.
4. **Resume orchestration script**: fetch latest global → patch job (`source_ckpt`, remaining rounds) → submit → monitor.

## Evidence
- `40c0948f`: round-10 global stranded on crashed UMCU_1.
- `696648fe`: died at round 7 — all compute lost.

## References
PR #345, `docs/SWARM_FAILURE_MODES.md`, runs `40c0948f` / `696648fe`.

Ultimate-Storm:
**Update from attempting warm-start in practice (June 2026)**

The warm-start *mechanism* works without any rebuild: `PTFileModelPersistor` accepts `source_ckpt_file_full_name`; a **relative** value resolves to `<app_root>/custom/`, and it round-trips a bare `FL_global_model.pt` state-dict (via `PTModelPersistenceFormatManager`). Both the config and the checkpoint ride with the submitted job (**byoc** — no kit rebuild).

[...]

Issue #346, opened 2026-06-19
Swarm: a single node's transient failure aborts the entire multi-hour run


Ultimate-Storm:
## Problem
In CCWF swarm runs, **any single client's fatal error aborts the entire run**, discarding all completed rounds. This happens both when a client's trainer raises a `FATAL_SYSTEM_ERROR` and when a peer's contribution returns a comms `ERROR` during the rotating aggregator's round gather. With `min_responses_required = #clients` (wait-for-all) plus 24 h timeouts, there is **no tolerance for a transient single-node hiccup**.

## Evidence (8-site ODELIA experiment, June 2026)
- **Run `40c0948f`** (6 nodes, Jun 17–18): reached **round 10/20** after ~17.5 h, then **UMCU_1** sent a fatal failure at the round 10→11 transition → full abort.
- **Run `696648fe`** (7 nodes, Jun 18–19): reached **round 7/20** after ~15 h, then **VHIO_1**'s contribution returned `ERROR` during RSH's round-8 peer gather (root cause: a ~6-min VPN drop — see the VPN issue) → full abort.
- Server log signature in both: `Aborting current RUN due to FATAL_SYSTEM_ERROR received: received failure report from client X: ERROR`.

Over a 15–17 h, 7-node run the probability that *some* node hits a transient error is high, so this single-point-of-failure behavior has so far prevented any 20-round run from completing.

## Why current hardening doesn't cover it
The 24 h timeouts (PR #345) protect against slow/silent nodes (F1), and wait-for-all fixes the `MODEL_UNRECOGNIZED` desync (F5). But an explicit `ERROR` return is treated as **immediately fatal** — it is not a timeout, so the long timeouts never apply.

## Proposed fixes (options, not mutually exclusive)
1. **Tolerate a transient peer failure in the gather**: retry the learn request to a peer that returns a comms `ERROR` (e.g. `target_unreachable` from a brief VPN blip) before declaring failure — only fatal after N retries / a grace window.
2. **Configurable per-round fault tolerance**: allow a round to proceed with `min_responses_required` successful contributions and *drop* a node that errored that round (re-include it next round), instead of aborting. Requires the aggregator to reject only the stale/errored update, not the whole run (mind the F5 desync trade-off).
3. **Checkpoint + auto-resume** so an unavoidable abort doesn't lose progress (see the warm-continue issue).
4. Investigate `SwarmClientController` / `SwarmServerController` knobs for treating a peer `ERROR` as recoverable.

[...]

Issue #160, opened 2025-11-24
Add mechanism to collect log files and checkpoints from swarm


SteffenRenisch:
During the debugging / investigation phase, we need to collect the logfiles from all swarm participants. Think about using the NVFlare messaging mechanism to collect and consistently store the logfiles per experiment.

Ultimate-Storm:
Possible solutions:

- [ ] NVFlare messaging mechanism
- [ ] As a job

SteffenRenisch:
Also collect files containing GT/class probabilities (#177 ) from sites for retrospective evaluation.

oleschwen:
If we want to collect any checkpoints (e.g., for evaluation on the held-out test data), decide which checkpoints to collect and do so via the same mechanism.
Maybe the server could already store them during swarm training (if possible).

oleschwen:
First step: transfer log file from each client to the server at the end of the training. @SteffenRenisch looks into how this can be done via NVFlare mechanisms.

oleschwen:
Locations of files to be collected are documented in assets/readme/README.participant.md

Ultimate-Storm:
# Live log and artifact collection for provisioned startup kits

## Overview

Provisioned startup kits can be patched so that log and artifact collection starts automatically when collaborators launch either:

```bash
./docker.sh --data_dir $DATADIR --scratch_dir $SCRATCHDIR --GPU device=0 --local_training 2>&1 | tee local_training_console_output.txt
```

or:

```bash
./docker.sh --data_dir $DATADIR --scratch_dir $SCRATCHDIR --GPU device=0 --start_client
```

The collection runs in parallel with training and client execution and continuously uploads files to the central server through SSH keys.

Current remote upload target:

```text
mediswarm-upload@172.24.4.65:/srv/mediswarm/live
```

## What is collected

### Local training mode

From the startup kit directory:

- `startup/local_training_console_output.txt`

From the local run directory:

- `startup/runs/$SITE_NAME/<RUN_NAME>/last.ckpt`
- `startup/runs/$SITE_NAME/<RUN_NAME>/epoch=*.ckpt`
- `startup/runs/$SITE_NAME/<RUN_NAME>/lightning_logs`
- `startup/runs/$SITE_NAME/<RUN_NAME>/*_model_gt_and_classprob_*.csv`

### Swarm client mode

From the startup kit:

- `startup/nohup.out`
- `log.txt`

From the job directory:

[...]

PR #357, opened 2026-06-24
Warm-continue + central checkpointing for swarm runs (#347, #160)

Ultimate-Storm:
Refs #347, #160. **Stacked on #346** (chain: main ← #345 ← #346 ← #347; retarget to `main` as the chain merges).

Solves losing all progress on an abort, and the 689 MB warm-start *delivery* problem (bundling overruns the deploy timeout).

- **`WarmStartablePTFileModelPersistor`** (`_shared/custom/warm_continue.py`): mirrors each new best global to a **uniform, run-independent local path** (`/scratch/mediswarm_latest_global.pt`, persists on the scratch mount across runs) and **gracefully inits fresh** if `source_ckpt` is absent (first run).
- **Configs**: persistor → the subclass with `source_ckpt_file_full_name` = the uniform path → a chain of short runs **auto-resumes** (each block continues from the prior block's global, read locally — no bundling, no per-site staging).
- **`scripts/collect_swarm_globals.sh`** (#160): pull the mirrored global + client log from every site to a central dir, so progress is never stranded on a crashed node.

**Testing:** verify the mirror is written each round, that a fresh first run starts clean, and that a second run warm-starts from the mirror; then chain short blocks and confirm rounds accumulate. Eval the result vs current best (macro 0.677 / class-2 0.837).

### DIFF
--- a/application/jobs/ODELIA_ternary_classification/app/config/config_fed_client.conf
+++ b/application/jobs/ODELIA_ternary_classification/app/config/config_fed_client.conf
@@ -108,8 +108,13 @@
     }
     {
       id = "persistor"
-      path = "nvflare.app_opt.pt.file_model_persistor.PTFileModelPersistor"
+      path = "warm_continue.WarmStartablePTFileModelPersistor"
       args {
+        # WARM-CONTINUE (#347): warm-start from the prior run's global mirrored at this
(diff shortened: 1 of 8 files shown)


oleschwen (approved):
Looks good for an experimental run in the swarm, but we may want to reconsider the global path (e.g., as part of #347, which is not closed by this PR)

Issue #316, opened 2026-04-30
Scaling readiness: 3 → 8 collaborators for full ODELIA swarm experiment


Ultimate-Storm:
## Context

The full ODELIA swarm experiment will scale from 3 nodes (PR301 validation deployment) to 8 collaborators with substantially larger combined datasets. The MST validation already showed that round time is governed by the slowest site and dominated by per-round prediction export and live-sync I/O. Both will get worse at 8x and with bigger checkpoints (`1DivideAndConquer`).

## Pitfalls to address before launch

1. **Per-round aggregated prediction export** — see linked issue. Must be throttled to final-round-only (or every Nth round) before launch, otherwise round time blows up.
2. **DataLoader worker cap** — see linked issue. Per-node tuning needed.
3. **Live-sync artifact scanning at the slowest node** — file count grows with rounds × sites × artifact types. Verify the scan/upload path doesn't pile up.
4. **Model transfer + fd pressure with `1DivideAndConquer`** — checkpoints are larger; PR312's `nofile=65536` raise should hold, but verify under 8-collaborator load.
5. **Server-side aggregation cost / memory** — confirm aggregator handles 8 client deltas of the larger model without OOM.
6. **Deploy tooling** — `run_deploy_test.sh` and `run_duke_iid_startupkit_smoke.sh` are written for 3 sites; confirm they handle N sites or extend.
7. **Stale-container cleanup** — PR312 broadened cleanup; verify it scales when many sites share a host.

## Acceptance

- Pre-launch checklist completed and signed off.
- Dry run on a representative subset (e.g. 5 nodes) before the full 8-node experiment.

## Related

- #301, #312
- Linked: prediction export throttle, worker tuning, 1DivideAndConquer validation issues (above).

Ultimate-Storm:
### Status consolidation (2026-07-02): implementation of the robustness pitfalls is done; this issue now tracks the remaining **live at-scale validation**

Closing #346, #347, #348, #315 as **implemented + merged** (validated at 2–3 sites where possible). Their residual acceptance is a live 8-node (or 5-node dry-run) window, which is this issue’s purpose — consolidated here so nothing is lost.

**Landed on `main` (implementation complete):**
[...]

PR #358, opened 2026-06-25
Throttle per-round aggregated prediction export in swarm mode (#314)

### LINKED ISSUE #314: Throttle per-round aggregated prediction export in swarm validate_and_train
## Context

The PR301 MST validation run on the 3-node startup kit was about 3.3x slower than the Apr 26 baseline (median round interval ~34 min vs ~10 min). The bottleneck is the per-round aggregated prediction export performed inside `validate_and_train()` in [application/jobs/_shared/custom/threedcnn_ptl.py:587-608](application/jobs/_shared/custom/threedcnn_ptl.py#L587-L608).

Each round, after validation, the swarm code calls `output_GT_and_classprobs_csv()` ([threedcnn_ptl.py:429-465](application/jobs/_shared/custom/threedcnn_ptl.py#L429-L465)) twice — once over the full train dataloader and once over the full val dataloader — running per-sample inference (`batch_size=1`) and appending rows to:

- `aggregated_model_gt_and_classprob_train.csv`
- `aggregated_model_gt_and_classprob_validation.csv`

In swarm mode this is enabled by default (see [threedcnn_ptl.py:63](application/jobs/_shared/custom/threedcnn_ptl.py#L63)); preflight/local mode already passes `output_GT_and_classprob=False` ([threedcnn_ptl.py:66](application/jobs/_shared/custom/threedcnn_ptl.py#L66)).

Cost scales linearly with `|train| + |val|` and with the slowest node, and currently runs every single round.

## Why it matters before scaling 3 → 8 nodes

Round time is governed by the slowest collaborator. With 8 sites and larger datasets, full-dataset inference + CSV writing every round will dominate wall-clock time and won't be informative more than once or twice.

## Proposal

Make per-round aggregated prediction export configurable in swarm mode:

- Default to **final-round only** (or every Nth round), not every round.
- Expose via env var (e.g. `ODELIA_PRED_EXPORT_EVERY=N`, `0` = final round only) and threaded through to `validate_and_train()`.
- Keep current behavior available for debugging.
- Confirm `GT_PredProb_Output_Callback` (the per-epoch local callback at [threedcnn_ptl.py:476-481](application/jobs/_shared/custom/threedcnn_ptl.py#L476-L481)) is left untouched or also gated — it currently fires every local epoch and contributes to the same CSV write pressure.

## Acceptance

- Configurable cadence with documented default.
- Round-interval regression closes most of the ~24 min gap on the 3-node MST run.
- 8-node projection is within tolerance for the planned full-scale experiment.

Ultimate-Storm:
Closes #314.

The per-round aggregated prediction export in `validate_and_train()` (`output_GT_and_classprobs_csv` — full train+val inference at `batch_size=1` + CSV writes, `threedcnn_ptl.py`) ran **every round** in swarm mode and **dominated round time** (PR301 validation was ~3.3× slower because of it), while being uninformative more than once or twice — a key blocker for the 3→8 node scale-up (#316).

**Change:** gate the export in the swarm loop (`main.py`) via a new helper `threedcnn_ptl.should_export_aggregated_predictions(current_round, total_rounds)`, controlled by **`ODELIA_PREDICTION_EXPORT_EVERY_N_ROUNDS`**:
- `0`/unset → **final round only** (new default)
- `1` → every round (legacy behaviour)
- `N` → every Nth round + the final round

`current_round`/`total_rounds` come from the `FLModel` received each round. `validate_and_train` itself is unchanged (still parameterised by `output_GT_and_classprob`). Documented in `README.developer.md`.

**Verification:** `py_compile` clean; the cadence logic unit-tested (15/15 cases: default=final-only, every-round, every-Nth+final, unknown-total, bad-value fallback). Behaviour-preserving for local/preflight (already passed `output_GT_and_classprob=False`).

### DIFF
--- a/application/jobs/_shared/custom/env_config.py
+++ b/application/jobs/_shared/custom/env_config.py
@@ -15,7 +15,12 @@ def _env_flag(name: str, default: bool = False) -> bool:
 
 def load_environment_variables():
     scratch_dir = os.environ['SCRATCH_DIR']
-    default_loader_workers = min(mp.cpu_count(), 8)
+    # Default DataLoader worker count. PR301 capped this at 8, which regressed
+    # loader concurrency on the ~16-CPU deploy nodes (the older path used all
+    # CPUs). Raise the cap to 16 so those nodes are not throttled by default,
+    # while still bounding very-large hosts. Tune per node via ODELIA_NUM_WORKERS
+    # (lower it on RAM-constrained nodes). See #315.
+    default_loader_workers = min(mp.cpu_count(), 16)
 
     return {
         'site_name': os.environ['SITE_NAME'],
--- a/application/jobs/_shared/custom/main.py
+++ b/application/jobs/_shared/custom/main.py
@@ -58,9 +58,19 @@ def main():
 
             while flare.is_running():
                 input_model = flare.receive()
-                logger.info(f"Current round: {input_model.current_round}")
+                current_round = input_model.current_round
+                total_rounds = getattr(input_model, "total_rounds", None)
+                logger.info(f"Current round: {current_round} (of {total_rounds})")
 
-                threedcnn_ptl.validate_and_train(logger, data_module, model, trainer, path_run_dir)
+                # #314: the per-round aggregated prediction export is expensive and
+                # dominates swarm round time; throttle it (default: final round only).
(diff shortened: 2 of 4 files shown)


oleschwen (approved):
looks good

PR #360, opened 2026-06-25
GPU-container watchdog: restart a GPU-less idle client unless training (#343)

### LINKED ISSUE #343: Address Docker Containers losing GPU
Swarm client containers sometimes lose access to their GPU. Running training jobs continue, but newly started ones fail due to lack of GPU.
#245 introduced a health check using `nvidia-smi` and a more elaborate health check script.
If we cannot prevent containers from losing their GPU, we probably want to restart such containers _unless_ they are running a training job.

Ultimate-Storm:
Closes #343.

When a client container loses its GPU (cgroup v2 + `systemd` driver + daily `daemon-reload`), its `nvidia-smi` HEALTHCHECK marks it **`unhealthy`**: a job that already grabbed the GPU keeps running, but newly started ones fail. The durable fix is the `cgroupfs` driver (`fix_docker_cgroupfs.sh`, from #345); this PR adds the **fallback** the issue asks for — restart such containers *unless they are running a job*.

**`scripts/client_node_setup/gpu_container_watchdog.sh`** (systemd timer, every 2 min):
- container `healthy`/starting → no-op.
- `unhealthy` **and** a GPU compute process is running on the host → a job grabbed the GPU before the loss → **leave it** (restarting would kill the job).
- `unhealthy` **and** no GPU compute process → idle → `docker restart` (re-applies the device cgroup, restoring the GPU).

Reuses the existing client HEALTHCHECK (`master_template.yml`) for GPU-loss detection and the host `nvidia-smi --query-compute-apps` to distinguish idle from active. Documented as the F2 fallback in `SWARM_FAILURE_MODES.md`. `bash -n` clean. *(The preflight already `[WARN]`s on the systemd driver and `[FAIL]`s on actual GPU-loss-in-container — left as-is so this watchdog can still rescue systemd hosts.)*

### DIFF
--- a/docs/SWARM_FAILURE_MODES.md
+++ b/docs/SWARM_FAILURE_MODES.md
@@ -35,6 +35,7 @@ Most of these are now caught automatically by the pre-run checks in the startup-
 - **Detection:** the pre-run check warns when the Docker cgroup driver is `systemd`; `docker exec <client> nvidia-smi -L` fails while host `nvidia-smi` works.
 - **Fix (host-level, durable):** run **`scripts/client_node_setup/fix_docker_cgroupfs.sh`** (as root) — switches Docker to the `cgroupfs` driver and validates that a `daemon-reload` no longer drops the GPU. Then recreate the client (see F3). The change persists across reboots.
 - **Prevention:** use the `cgroupfs` driver on all client hosts; the pre-run check flags `systemd`.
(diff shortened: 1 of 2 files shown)


oleschwen (approved):
Looks good.
We should check the effect in a productive swarm.

PR #366, opened 2026-06-29
363 Download ResNet18 pretrained weights during Docker build

### LINKED ISSUE #363: ODELIA CI pre-flight check should use pre-trained model from image
The CI preflight check uses ResNet18, downloading model weights during the test. Like the other models, we should provide pre-trained model weights via the Docker image to avoid depending on successful download during pre-flight check.

oleschwen:
This PR downloads the pre-trained weights for ResNet 18 when building the Docker image and copies them into the image.
When running, e.g., the data access preflight check, we still see a message that the weights are downloaded, but they are in fact taken from the existing cache.
Follow-up issue: #376


Ultimate-Storm (approved):
Not sure about hardcoded HF cache layout, but should be fine

PR #400, opened 2026-07-06
Deploy: --restart=unless-stopped for server/client containers

### LINKED ISSUE #393: Deploy: use --restart=unless-stopped for server/client containers (survive host reboots)
**Priority: HIGH (quick win)** · Area: deploy / kit template

## Problem
The kit-generation template launches both the server and client containers with `--restart=on-failure:5`:
- `docker_config/master_template.yml:1013` (server)
- `docker_config/master_template.yml:1093` (client)

A **host reboot is not a container "failure"**, so `on-failure` does not bring the container back — it stays stopped. This is the **exact root cause of the 2026-07-05 weekend outage**: the Cosmos host auto-rebooted (unattended-upgrades) and the FL server never restarted, so the swarm was dead until manually recovered with `docker update --restart always`. The same applies to every site's client.

This template is the authoritative one — the build copies it over the fork's copy at `docker_config/Dockerfile_ODELIA:267`.

## Fix
- [ ] Change `--restart=on-failure:5` → `--restart=unless-stopped` at `master_template.yml:1013` (server) and `:1093` (client). (`unless-stopped` restarts on reboot but respects a deliberate `docker stop`; `always` also works but restarts even after intentional stops.)
- [ ] Rebuild the image and regenerate startup kits so new kits carry the policy.
- [ ] Update operator docs (`assets/readme/README.operator.md`, `docs/ROLLOUT_v1.5.0.md`) to drop the manual `docker update --restart always` step.
- [ ] Note in the changelog that existing/deployed containers still need a one-time `docker update --restart unless-stopped` (policy change only applies to newly-created containers).

## Verification
- [ ] On a test host, start a kit container, `sudo reboot`, and confirm the container is `Up` automatically afterward.
- [ ] `docker inspect -f '{{.HostConfig.RestartPolicy.Name}}' <container>` reports `unless-stopped`.

Ultimate-Storm:
Closes #393.

Changes `--restart=on-failure:5` -> `--restart=unless-stopped` for both the server (master_template.yml:1093) and client (:1013), so containers auto-restart after a host reboot — the root cause of the 2026-07-05 outage (on-failure does not survive a reboot). Updates the ROLLOUT runbook.

**Verification:** YAML parses. Takes effect on newly-built kits; existing containers need a one-time `docker update --restart unless-stopped`. A reboot smoke-test is deferred to the build/deploy window (Cosmos/dl occupied by STAMP eval).

### DIFF
--- a/docker_config/master_template.yml
+++ b/docker_config/master_template.yml
@@ -1010,7 +1010,7 @@ docker_cln_sh: |
 
   elif [ -n "$START_CLIENT" ]; then
       _start_live_sync swarm
-      docker run -d -t --restart=on-failure:5 \
+      docker run -d -t --restart=unless-stopped \
       --health-cmd="nvidia-smi > /dev/null 2>&1 || exit 1" \
       --health-interval=120s --health-start-period=180s --health-retries=3 \
       $DOCKER_OPTIONS $ENV_VARS --env TRAINING_MODE=swarm $DOCKER_IMAGE \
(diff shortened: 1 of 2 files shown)


oleschwen (approved):
looks good

Issue #396, opened 2026-07-06
CI: pr-test / unit-test efficiency (Docker + pip caching, parallelize GPU steps)


Ultimate-Storm:
**Priority: MEDIUM** · Area: CI/CD

## Problem
`.github/workflows/pr-test.yaml` (self-hosted GPU, 60-min timeout) has avoidable inefficiency:
- The **"Build Docker image for real project (MEVIS)"** step calls `buildDockerImageAndStartupKits.sh` **without `--use-docker-cache`**, while the very next dummy-project and STAMP builds both pass `--use-docker-cache`. The uncached MEVIS build likely rebuilds the 5.2 GB image from scratch every PR run.
- It then runs **13 sequential** `runIntegrationTests.sh` steps (dummy standalone/simulation/poc/swarm, 3dcnn, STAMP preflight/local/simulation, startup kits, licenses, GPU preflight, data preflight) on one job — they run back-to-back even where independent.
- `.github/workflows/unit-tests.yaml` installs torch + lightning every run with **no pip cache**.

## Fix
- [ ] Add `--use-docker-cache` to the MEVIS build step (consistent with the others), or build the image **once** and reuse the tag across subsequent steps.
- [ ] Group/parallelize independent integration steps — e.g. split the **STAMP** suite and the **ODELIA** suite into separate jobs, or matrix them — respecting the existing `mediswarm-self-hosted-gpu-validation` single-GPU concurrency guard (serialize only the GPU-bound steps).
- [ ] Add `actions/setup-python` `cache: pip` (or `actions/cache` on the pip dir) to `unit-tests.yaml`.
- [ ] Optionally collapse the 13 `continue-on-error: false` steps into fewer grouped invocations to cut per-step runner overhead.

## Verification
- [ ] PR-validation wall-clock time drops measurably (compare a before/after run).
- [ ] The MEVIS build reuses cached layers (build log shows `CACHED`).
- [ ] `unit-tests.yaml` shows a pip cache hit on the second run.

oleschwen:
We could also consider using a persistent cache of downloaded pre-trained model weights by setting MEDISWARM_BUILD_CACHE_DIR to a suitable location.

Ultimate-Storm:
**Partially done by #401** (merged):
- ✅ MEVIS image build now uses `--use-docker-cache` (it was the only build rebuilding 5.2 GB from scratch every PR)
- ✅ `unit-tests.yaml` pip caching

**Still open — the bigger half:**
[...]

PR #401, opened 2026-07-06
CI: cache MEVIS image build + unit-test pip deps

Ultimate-Storm:
Part of #396.

- pr-test.yaml: MEVIS build now passes `--use-docker-cache` (matches dummy/STAMP builds) instead of a full 5.2 GB rebuild.
- unit-tests.yaml: enable actions/setup-python pip caching.

GPU integration-step parallelization is left as a follow-up on the same issue. **Verification:** both workflows parse; CI-run timing comparison deferred to a runner.


oleschwen (approved):
looks good

Issue #398, opened 2026-07-06
Observability: harden live-sync + reliable swarm round-progress telemetry


Ultimate-Storm:
**Priority: MEDIUM-HIGH** · Area: application / observability

## Problem
During the 2026-07-06 swarm run, round progress was effectively **invisible**:
- The CCWF swarm workflow hands off at `swarm_start` and then runs **all rounds peer-to-peer on the clients** ("Waiting for clients to finish workflow"), so the **server log is quiet by design** — round detail only appears as sparse status pushes.
- The `server_tools/app.py` live-monitor (2072-line Flask app) depends on per-site `live_sync.sh` uploads, but **heartbeats were stale**: `kit_live_sync/live_sync.sh` runs under `set -euo pipefail` and **dies when its `.mediswarm_sync` files are root-owned** (which happens when a client was started via `sudo`). Once the daemon dies, that site stops reporting.

Net effect: I had to `grep nohup.out` by hand to tell whether a round advanced, stalled, or aborted.

## Fix
- [ ] Harden `kit_live_sync/live_sync.sh`: create/own its state files as the **site user**, and **don't abort the whole daemon** on a single `touch`/upload failure (guard the `set -e` sections; log-and-continue). Detect/repair root-owned `.mediswarm_sync`.
- [ ] Emit a **structured per-round progress event** (round number, per-client `start/finished_learn`, prunes) that the monitor can read even while a round is mid-flight — so a stall vs. slow-training is distinguishable without manual log grepping.
- [ ] Surface prune/abort/stall signals in `server_tools/app.py` prominently (badge per site + last-round timestamp).
- [ ] Cross-link the failure modes in [[odelia-swarm-debugging]].

## Verification
- [ ] Start a client via `sudo`, run a job, and confirm live-sync keeps reporting (no death on root-owned files).
- [ ] The monitor shows the current round advancing during a live run without reading `nohup.out`.
- [ ] A stalled round is visibly distinguishable from slow training in the monitor.

PR #403, opened 2026-07-06
Observability: harden live_sync.sh against root-owned state dir

Ultimate-Storm:
Part of #398.

Makes `live_sync.sh` survive a root-owned `.mediswarm_sync` (from a sudo-launched client) instead of dying under `set -euo pipefail` and going silent — the cause of stale live-monitor heartbeats on 2026-07-06. Reclaims ownership via passwordless sudo, else falls back to a user-writable dir; guards the timestamp writes.

Follow-up (same issue, deferred — needs a live run): structured per-round progress telemetry in server_tools/app.py. **Verification:** `bash -n`.

### DIFF
--- a/.github/workflows/unit-tests.yaml
+++ b/.github/workflows/unit-tests.yaml
@@ -41,6 +41,8 @@ jobs:
           # Install unified 'lightning' package — provides both lightning.*
           # (needed by STAMP tests) and pytorch_lightning.* (needed by ODELIA tests)
           pip install lightning scikit-learn pytest
+          # fastapi: server_tools/app.py imports it; needed by test_swarm_progress.py (#398)
+          pip install fastapi
 
       - name: Run unit tests
         run: |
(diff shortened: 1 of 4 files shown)


Ultimate-Storm:
**Live-validated on a real 2-client swarm (2026-07-07 diagnostic).** Pointed the round-progress panel at the running server `nohup.out`: it correctly showed job / round 1/3 / status — and this caught a real bug: the per-client chips were **empty for lowercase client names** (`node_A`/`node_B`) because the regex only matched uppercase (`USZ_1`). Fixed both regexes to `[A-Za-z0-9_]+`, added `tests/unit_tests/test_swarm_progress.py` (uppercase + lowercase + stall + idle) and the `fastapi` unit-test dep. Re-validated against the live run (renders the `node_B` chip) + all unit tests green.

oleschwen (approved):
looks good

PR #414, opened 2026-07-09
351 ODELIA CI should use unproblematic datasets

### LINKED ISSUE #351: ODELIA CI CNN training should use unproblematic dataset
Following #349:
The CI tests training productive models currently use the synthetic dataset with implausibilities (intended for the test whether such implausibilities are caught). Instead, a "clean" dataset should be used.

oleschwen:
This PR adds a separate test client node with a problematic dataset, used only in the tests whether problematic datasets are handled correctly.
Other tests now use unproblematic datasets, avoiding unnecessary and confusing warnings.


Ultimate-Storm (approved):
scripts and tests, approved with CI green

PR #415, opened 2026-07-09
370 more consistent hash function

### LINKED ISSUE #370: Consider using consistent hash function
We're currently using at least three different hash functions (md5sum when restarting training, sha (i.e., sha1) for verifying pre-trained weight model files, and sha3_224 for reporting image duplicates. Not a problem per se, but we also have no specific requirements why we want specifically those, computing hashes should not be a performance bottleneck anywhere, nor should memory due to hash length.
I would suggest to use a consistent hash function everywhere. sha_224 or sha3_224 should be long enough for our purposes and we would (unlike for md5sum and sha1) not have to think about collision attacks at the moment.

oleschwen:
This PR unifies hash functions used to 256-bit SHA2.


Ultimate-Storm (approved):
approved

PR #417, opened 2026-07-09
STAMP: read features extracted by STAMP 2.5.0 (raise readable version ceiling)

Ultimate-Storm:
> **Stacks on #413** (the `--env NAME` space-forwarding fix). Once #413 merges this
> PR's diff reduces to the version-gate commit alone.

## Problem

UKB Bonn extracted DECADE features with **STAMP 2.5.0**. Our training image runs
STAMP 2.4.0, which refuses them:

```
RuntimeError: features were extracted with a newer version of stamp,
please update your stamp to at least version 2.5.0.
```

Simply bumping the pin is not possible: **STAMP 2.5.0 hard-requires Python 3.13**
(`requires-python = ">=3.13,<3.14"`), while the image runs Python 3.11 and our
NVFlare fork declares support only through 3.12. Migrating the whole stack to
3.13 days before the first consortium run is not warranted.

## Why reading 2.5.0 features on 2.4.0 is safe

The check is a conservative forward-compat guard, not a format break. Between the
`2.4.0` and `2.5.0` tags of KatherLab/STAMP:

* the feature H5 layout is **identical** — datasets `feats`/`coords`, attrs
  `stamp_version, extractor, unit, tile_size_um, tile_size_px, code_hash, feat_type`;
* `preprocessing/extractor/uni.py` is **byte-identical** → UNI features are
  numerically the same;
* the only `preprocessing/tiling.py` changes are bug fixes (eager PIL decode; a
  None-guard on MPP metadata) — they do not alter tiles, coords or features.

Empirically, the loader already parses these files (`Loaded 15 patients`) and only
then trips the guard.

## Change

`scripts/build/patch_stamp_feature_version_gate.py`, applied at image build, raises
only the **readable** extraction-version ceiling to 2.5.0:

```python
) > max(Version(stamp.__version__), Version("2.5.0")):
```

Anything newer still raises. The patch **fails the build loudly** if the upstream
guard is not found, so a future STAMP pin bump cannot silently skip it. It is
idempotent.

Docs record the DECADE policy: sites extract with STAMP 2.5.0.

## Verification (synthetic data, RTX 6000 clients)

| Case | Result |
|---|---|
| features tagged `2.5.0` | preflight **exit 0**, `Loaded 15 patients` |
| features tagged `2.6.0` | **still raises** (guard intact) |
| `2.5.0` features + `Slide ID` column (Bonn's exact case) | **exit 0** |
| partner smoke: dummy / plausibility / local on 2.5.0 features | all pass |
| 2-node deploy test (swarm + eval + compare-local) on 2.5.0 features | see below |

### DIFF
--- a/assets/readme/README.participant.STAMP.md
+++ b/assets/readme/README.participant.STAMP.md
@@ -84,6 +84,14 @@ Coordinate with your swarm operator to ensure all sites use the **same feature
 extractor, tile size, and feature dimension** — mismatched features will cause
 training failures.
 
+**STAMP version for extraction (DECADE): 2.5.0.** Extract with standalone STAMP
+2.5.0 (it requires Python 3.13). The swarm training image itself runs STAMP
+2.4.0 — because NVFlare/PyTorch in the image are on Python 3.11 — but it is
+patched to read features extracted by STAMP up to 2.5.0. The feature H5 layout
+and the UNI extractor are identical between 2.4.0 and 2.5.0, so this is safe.
+If you extract with a STAMP newer than 2.5.0, training will refuse the features
+— tell your swarm operator first.
+
 ### Folder Structure
 
 ```
--- a/docker_config/Dockerfile_STAMP
+++ b/docker_config/Dockerfile_STAMP
@@ -48,8 +48,21 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
 # This pulls in all STAMP dependencies: h5py, lightning, timm, transformers,
 # opencv-python, openslide-bin, openslide-python, lifelines, beartype, etc.
 # Using --no-cache-dir to keep the layer small.
+#
+# NOTE: we deliberately stay on 2.4.0. STAMP 2.5.0 hard-requires Python 3.13
+# (requires-python = ">=3.13,<3.14"), while this image runs Python 3.11 and our
+# NVFlare fork declares support only through 3.12. Sites extract features with
+# STAMP 2.5.0; the patch below lets this loader read them (the H5 layout and the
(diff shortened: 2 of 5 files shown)


Ultimate-Storm:
### Deploy-test validation (2-node, features tagged `stamp_version=2.5.0`)

Ran the full 2-node deploy test on `jefftud/decade:1.5.0-dev.260709.8afff27`
(UKB_1 on dl0, Mainz_1 on dl2), with every site's H5 tagged as extracted by
STAMP 2.5.0 — a faithful rehearsal of the real DECADE setup:

```
training_status  : pass
evaluation_status: pass
  global Mainz_1 -> acc 1.0  auroc 1.0
  global UKB_1   -> acc 1.0  auroc 1.0
  baseline       -> acc 0.75 auroc 1.0
federated_vs_local: {"comparable": true, "federated": 1.0, "local": 1.0,
                     "federated_at_least_as_good": true, "regression_detected": false}
```

Plus, on the same image: partner smoke (dummy / plausibility / local) on 2.5.0
features all pass; `2.6.0`-tagged features still raise (guard intact).

---

**Unrelated operational finding while validating** (not fixed here): the admin
container on the server host could not resolve `dl3.tud.de`, because
`fix_remote_dns()` in `run_stamp_deploy_test.sh` skips the local/Cosmos host. The
submit step then printed `[OK] Job submitted` even though `expect` had failed with
`FLCommunicationError: cannot connect to server`, and the run hung for the full
timeout. Two hardening items worth a follow-up:

1. ensure `dl3.tud.de` resolves on the admin/server host too (not just clients);
2. verify the job id returned by `submit_job` instead of unconditionally printing `[OK] Job submitted`.

Issue #419, opened 2026-07-09
Security: mediswarm-upload authorized_keys are unrestricted — partner sites get an interactive shell on the monitor/FL-server host


Ultimate-Storm:
## Problem

The `mediswarm-upload` account exists **solely** so partner sites can rsync live-monitor logs into `/srv/mediswarm/live/`. But:

- its login shell is **`/bin/bash`**, and
- **none** of the ~16 `authorized_keys` entries carry `command=` or `restrict`.

So **any partner site holding an upload key can obtain a full interactive shell** on the monitor host — which is also the **production FL server** and runs the Kather-portal stack.

**Confirmed in the field:** a partner's own screenshot (RUMC, 2026-07-09) shows them sitting at `mediswarm-upload@agh1:~$` running arbitrary commands.

## Fix — forced command + `restrict` on every key

Allow only what `live_sync.sh` actually needs, and nothing else:

| Needed | Source |
|---|---|
| `rsync --server …` (upload) | `rsync_cmd` |
| `mkdir -p '<path under /srv/mediswarm/live>'` | `ensure_remote_dir` |
| `echo ok` | connectivity test (live_sync + partner instructions) |

- [ ] Install a guard script as the **forced command**; validate `$SSH_ORIGINAL_COMMAND` against the three shapes above, confine rsync to `/srv/mediswarm/live` (e.g. `rrsync -wo`), reject + **log** everything else.
- [ ] Prefix every key with `restrict,command="…"` (`restrict` implies `no-pty`, `no-port-forwarding`, `no-agent-forwarding`, `no-X11-forwarding`, `no-user-rc`).
- [ ] Keep comments/blank lines intact; back up `authorized_keys` first.
- [ ] Consider a dedicated group + tighter home-dir perms.

⚠️ **Roll out carefully:** a wrong forced command breaks log upload for *every* site. Back up, apply, then verify end-to-end before/after.

## Verification
- [ ] `ssh mediswarm-upload@host` (interactive) → **denied** (no shell).
- [ ] `ssh mediswarm-upload@host 'echo ok'` → `ok`.
- [ ] `ssh mediswarm-upload@host "mkdir -p '/srv/mediswarm/live/X/swarm/run'"` → succeeds.
- [ ] `rsync -az f mediswarm-upload@host:/srv/mediswarm/live/X/swarm/run/` → succeeds.
- [ ] `ssh mediswarm-upload@host 'cat /etc/passwd'` → **denied + logged**.
[...]

PR #420, opened 2026-07-09
Security: mediswarm-upload forced-command guard (#419)

Ultimate-Storm:
Adds `server_tools/mediswarm_upload_guard.sh` — the reproducible copy of the forced-command guard already installed + verified on the live monitor host (see #419). Restricts every `mediswarm-upload` key to `echo ok` / validated `mkdir` under /srv/mediswarm/live / `rsync --server`→`rrsync -wo /` (write-only). Verified end-to-end: uploads land, interactive shell + arbitrary commands + /etc/passwd exfil all blocked. Header documents install + authorized_keys prefixing. Closes the first checklist item of #419.

### DIFF
--- a/server_tools/mediswarm_upload_guard.sh
+++ b/server_tools/mediswarm_upload_guard.sh
@@ -53,7 +53,15 @@ fi
 #    filters dangerous rsync options (e.g. --rsync-path).
 if [[ "$cmd" == "rsync --server "* ]]; then
     log allow-rsync
-    exec /usr/bin/rrsync -wo /
+    # -no-lock: rrsync otherwise takes an exclusive flock() on the restricted root
+    # (here "/"), which serialises uploads across the WHOLE account. With several
+    # sites live-syncing every 30s, the first site's daemon holds the lock and every
+    # other site is refused with
+    #     rrsync error: Another instance of rrsync is already accessing this directory.
+    # Observed in DECADE: one site's daemon blocked uploads for all the others, so we
+    # were blind during the run. Each site writes into its own subdirectory, so the
+    # single-run lock buys us nothing.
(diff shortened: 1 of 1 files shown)


oleschwen (approved):
looks good
orchestration script only, MediSwarm and application code unaffected

Ultimate-Storm:
### `rrsync` locks the restricted root — one site blocks all the others

Pushed a fix to this branch. While validating the DECADE kits I found this guard
**breaks live-sync for the whole consortium**.

`rrsync` takes an **exclusive `flock()` on the restricted root** — which here is `/`.
So uploads are serialised across the entire `mediswarm-upload` account:

```
/usr/bin/rrsync error: Another instance of rrsync is already accessing this directory.
rsync: connection unexpectedly closed (0 bytes received so far) [sender]
```

Observed live today: UKB Bonn's live-sync daemon (`rrsync -wo /`, holding the lock on
`/`) made **every** upload from the other sites fail. A site sees
`SSH OK — live sync enabled` and then a stream of rsync errors — exactly what UM Mainz
reported. We would have been blind during the run: no logs, no heartbeats, no checkpoints.

Each site writes into its own subdirectory (`/srv/mediswarm/live/<SITE>/...`), so the
single-run lock buys nothing.

**Verified on a real client**, with a competing exclusive `flock` held on `/`:

| | before (`-wo /`) | after (`-wo -no-lock /`) |
|---|---|---|
| sequential upload ×3 | rc=12 (all fail) | rc=0 |
| two concurrent uploads | rc=12 / rc=12 | rc=0 / rc=0 |

Also applied to the deployed `/etc/ssh/mediswarm_upload_guard.sh` on the monitoring
host so the sites are unblocked right now.

PR #422, opened 2026-07-09
VPN: rewrite bare auth-user-pass so the systemd VPN service can auto-recover (#418)

### LINKED ISSUE #418: VPN: setup_vpntunnel.sh leaves a bare auth-user-pass, so the systemd VPN service prompts for credentials and cannot auto-recover
## Problem

`setup_vpntunnel.sh` writes the VPN credentials to `/etc/openvpn/client/mediswarm.auth` ([:84](scripts/client_node_setup/setup_vpntunnel.sh#L84)), then tries to point the config at them:

```bash
inject_if_missing() {
    if ! sudo grep -q "^$setting" "$file"; then  ...append...
    else echo "  Already present: $setting"; fi      # <-- skips
}
inject_if_missing "auth-user-pass" "auth-user-pass /etc/openvpn/client/mediswarm.auth"   # :128
```

GoodAccess `.ovpn` files ship a **bare `auth-user-pass`** (no path argument). `grep -q "^auth-user-pass"` therefore **matches**, the script prints `Already present: auth-user-pass`, and **never rewrites it to reference the credentials file**.

OpenVPN then prompts interactively, so the `mediswarm-vpn` systemd service **hangs at `Password entry required for 'Enter Auth Username'`** and can never start unattended.

**Observed live at RUMC (2026-07-09):** `Already present: auth-user-pass`, followed by two `systemd-tty-ask-password` broadcasts.

### Impact
This **silently defeats the auto-recovering VPN service (#348)** — the tunnel cannot come up unattended and will not reconnect after a drop. It is a strong candidate for **UKA's long-standing "silent VPN hang"**: the health watchdog restarts the service, and the service then waits forever for a password.

## Fix
- [ ] Distinguish a *bare* directive from one that already has a value:
  - `^auth-user-pass[[:space:]]+\S` → already configured, skip
  - `^auth-user-pass[[:space:]]*$` → **rewrite in place** to `auth-user-pass /etc/openvpn/client/mediswarm.auth`
  - absent → append
- [ ] Apply the same "present but valueless" handling to any other directive injected via `inject_if_missing`.
- [ ] After install, assert the service starts non-interactively (no password prompt) before reporting success.

## Verification
- [ ] With a GoodAccess `.ovpn` containing a bare `auth-user-pass`, `setup_vpntunnel.sh -d <site> -n -s` produces `/etc/openvpn/client/mediswarm.conf` containing `auth-user-pass /etc/openvpn/client/mediswarm.auth`.
- [ ] `systemctl start mediswarm-vpn` completes with **no** password prompt; tun IP appears; `systemctl is-active` = active.
- [ ] A `vpn_health_monitor` triggered restart also comes up unattended.
- [ ] Re-check **UKA** against this — likely the same root cause.

Ultimate-Storm:
Closes #418. GoodAccess .ovpn files ship a **bare** `auth-user-pass` (no path); the old `grep -q "^auth-user-pass"` matched it and skipped injecting the credentials path, so the `mediswarm-vpn` systemd service hung at *Enter Auth Username* — no unattended start / auto-recovery. Seen live at RUMC; **likely UKA's silent-VPN-hang root cause**. `ensure_auth_user_pass` keeps a path-form directive, rewrites a bare/comment one, else appends. Logic unit-tested across all 5 cases (bare / trailing-space / inline-comment / has-path / absent); `bash -n` clean. Deploy note: sites that ran the old script can repair `mediswarm.conf` by re-running `setup_vpntunnel.sh -d <site> -s`.

### DIFF
--- a/scripts/client_node_setup/setup_vpntunnel.sh
+++ b/scripts/client_node_setup/setup_vpntunnel.sh
@@ -124,8 +124,30 @@ if [[ -n "$SYSTEMD_MODE" ]]; then
         fi
     }
 
+    # auth-user-pass needs special handling: GoodAccess .ovpn files ship a *bare*
+    # `auth-user-pass` (no path), which makes OpenVPN prompt interactively so the
+    # systemd mediswarm-vpn service hangs at "Enter Auth Username" and can never
+    # auto-recover (#418). A plain grep "^auth-user-pass" matches the bare directive
+    # and skips injecting the path. So: keep it if it already has a path argument,
+    # rewrite a bare/comment one, otherwise append.
+    ensure_auth_user_pass() {
+        local file="/etc/openvpn/client/mediswarm.conf"
+        local authfile="/etc/openvpn/client/mediswarm.auth"
+        if sudo grep -qE "^auth-user-pass[[:space:]]+[^[:space:]#]" "$file"; then
+            echo "  Already present: auth-user-pass <path>"
+        elif sudo grep -qE "^auth-user-pass([[:space:]]*|[[:space:]]+#.*)$" "$file"; then
+            # delete-then-append avoids an s/// whose regex contains | (sed delimiter clash)
+            sudo sed -i -E "/^auth-user-pass([[:space:]]*|[[:space:]]+#.*)$/d" "$file"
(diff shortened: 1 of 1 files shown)


oleschwen (approved):
looks good
setup, does not affect MediSwarm or application code

Issue #274, opened 2026-04-08
STAMP: Test all model architectures in deploy test (MLP, TransMIL, Linear, Barspoon)


Ultimate-Storm:
## Context

The STAMP integration supports 5 model architectures:

| Model | \`STAMP_MODEL_NAME\` |
|-------|---------------------|
| VIT | \`vit\` |
| MLP | \`mlp\` |
| TransMIL | \`trans_mil\` |
| Linear | \`linear\` |
| Barspoon | \`barspoon\` |

The deploy test currently only validates **VIT**. Other architectures are tested in integration tests (simulator mode) but not in a real multi-node swarm deployment.

## Proposed Solution

Extend \`run_stamp_deploy_test.sh\` to iterate over models:

\`\`\`bash
STAMP_MODELS=("vit" "mlp" "trans_mil" "linear" "barspoon")
for model in "\${STAMP_MODELS[@]}"; do
    run_single_model "STAMP_classification" "\$model"
done
\`\`\`

This requires:
1. Exporting \`STAMP_MODEL_NAME\` dynamically per model before starting clients
2. Adjusting \`STAMP_DIM_INPUT\` if models expect different defaults
3. Potentially a \`--models\` CLI flag to select which models to test

## Related Files

- \`scripts/deploy/run_stamp_deploy_test.sh\` — hardcodes \`"vit"\` on line 752
- \`deploy_sites_2node_stamp_test.conf\` — site config
- \`application/jobs/STAMP_classification/app/custom/stamp_model_wrapper.py\` — model registry

Issue #270, opened 2026-04-08
STAMP: Add post-training evaluation/prediction step to deploy test


Ultimate-Storm:
## Context

The ODELIA deploy test (`run_deploy_test.sh`) has a full `evaluate_model()` step that:
1. Collects checkpoint files from remote client workspaces
2. Runs `predict.py` inside a Docker container to generate predictions
3. Validates model quality by checking prediction output

The STAMP deploy test (`run_stamp_deploy_test.sh`) currently has **no evaluation step** — it only checks that training completes without errors ("Server runner finished.").

## Proposed Solution

1. Implement a STAMP equivalent of `predict.py` (or a `stamp_predict.py`) that loads a trained STAMP model checkpoint and runs inference on held-out test data
2. Add `collect_checkpoints()` and `evaluate_model()` functions to `run_stamp_deploy_test.sh`
3. Optionally verify AUROC or other classification metrics against a minimum threshold on synthetic data

## Why This Matters

Without a prediction/evaluation step, the deploy test only validates that training **runs** — not that the resulting model is **usable**. This is a weaker guarantee than the ODELIA deploy test provides.

## Related Files

- `scripts/deploy/run_stamp_deploy_test.sh` — STAMP deploy test orchestrator (no evaluation step)
- `scripts/deploy/run_deploy_test.sh` — ODELIA deploy test (has `evaluate_model()` and `collect_checkpoints()`)
- `scripts/evaluation/predict.py` — ODELIA prediction script (no STAMP equivalent exists)
- `application/jobs/STAMP_classification/app/custom/stamp_training.py` — STAMP training code

Ultimate-Storm:
## Implementation approach (planned — deferred for now, for whoever picks this up)

Mirror the ODELIA deploy-test evaluation. ODELIA's `scripts/deploy/run_deploy_test.sh` has `collect_checkpoints()` (~L772) that pulls client checkpoints to a local staging dir and then runs `predict.py` in Docker; STAMP's `run_stamp_deploy_test.sh` currently stops at *"Wait for training completion"* (~L617) with no eval.

[...]

Issue #388, opened 2026-07-03
CI: validate-swarm flaky on the 3DCNN simulation/POC step (300s status timeout under load)


Ultimate-Storm:
## Problem

The `validate-swarm` CI job (`.github/workflows/pr-test.yaml`) intermittently fails on the 3DCNN dummy-swarm steps — **"Run dummy training in simulation mode"** and **"Run dummy training in proof-of-concept mode"** — even when the change under test cannot affect swarm training. This repeatedly forces admin-merges of otherwise-green PRs (docs, CI, STAMP, apt).

Observed on `main`/PRs on 2026-07-03 across several PRs, two distinct flake signatures:

1. **Simulation mode** (e.g. PR #387, run 28651372748 on runner `dl0`): the Docker build + training rounds 0–3 succeed, then
   ```
   ERROR - Aborting current RUN due to FATAL_SYSTEM_ERROR received:
           client simulated_node_1 didn't report status for 300 seconds
   ❌ Minimal example simulation mode failed.
   ```
2. **Proof-of-concept mode** (earlier PRs): `ValueError: I/O operation on closed file.` after `posix_spawn failed (setsid unavailable on this platform)` → `❌ Minimal example proof-of-concept mode failed.`

## Root cause (likely)

`tests/integration_tests/_run_3dcnn_simulation_mode.sh` pins **all** swarm timeouts to **300 s** (lines 31–43), including the client status-report timeout. On the shared self-hosted GPU runners (dl0/dl2) a round can stall > 300 s under GPU/IO contention → a healthy-but-slow `simulated_node_*` is declared dead → `FATAL_SYSTEM_ERROR` → whole-sim abort. This is the CI analogue of the production single-node-abort we already hardened (#346).

## Proposed fixes (options)

1. **Raise the status-report / progress timeouts** in `_run_3dcnn_simulation_mode.sh` from 300 s to something CI-load-tolerant (e.g. 900–1800 s). Cheapest, most direct.
2. **Automatic single retry** of the simulation/POC step in `runIntegrationTests.sh` before failing the job (a transient abort re-runs once).
3. Investigate the POC `I/O operation on closed file` / `posix_spawn setsid unavailable` path — that looks like an environment/teardown race distinct from the timeout, and may need its own fix.

## Acceptance
- `validate-swarm` passes reliably on unrelated (docs/CI/apt) PRs without admin override.
[...]

PR #424, opened 2026-07-09
STAMP deploy test: barspoon unsupported on 2.4.0 + fail the run when a model fails

Ultimate-Storm:
> **Stacks on #421 → #417 → #413.**

Found by actually running the deploy test across all five advertised STAMP models
(#274 covered exactly this and is *closed*).

## Bug 1 — `barspoon` cannot work on the shipped STAMP

`stamp.modeling.registry.ModelName` in STAMP **2.4.0** is `{vit, mlp, trans_mil, linear}`;
`models/barspoon.py` was **added in 2.5.0**. The client aborted with:

```
ValueError: failed to instantiate class: ValueError: 'barspoon' is not a valid ModelName
ServerRunner - ERROR - Aborting current RUN due to FATAL_SYSTEM_ERROR received:
  failed to configure clients ['UKB_1', 'Mainz_1']: only 0/2 configured
```

Yet `barspoon` was in the **default** model list, and both
`README.participant.STAMP.md` and the job README advertised it **to sites** — a
DECADE site selecting it would hard-fail. (Adopting STAMP 2.5.0 isn't an option
today: it requires Python 3.13; the image runs 3.11 and our NVFlare fork declares
support only through 3.12 — see #417.)

## Bug 2 — a failing model still exited 0

The run loop called `run_single_model` (which always `return 0`) and then ended.
No aggregate status, no exit code. That is how CI stayed green while barspoon was
broken:

```
[ERROR] FAILED: barspoon in 426s
DEPLOY_EXIT_CODE=0        # <-- reported success
```

## Changes

* Drop `barspoon` from `ALL_STAMP_MODELS`; reject it **early** with a message naming
  the required STAMP version, instead of aborting a swarm run 7 minutes in.
* Track `PASSED_MODELS` / `FAILED_MODELS`, print a summary, and **exit non-zero**
  when any model failed.
* Fix the model tables in the participant guide and job README.

## Evidence (2-node deploy test, features tagged `stamp_version=2.5.0`)

| model | training | evaluation | federated ≥ local |
|---|---|---|---|
| vit | pass | pass | ✅ |
| mlp | pass | pass | ✅ |
| trans_mil | pass | pass | ✅ |
| linear | pass | pass | ✅ |
| barspoon | **fail** | skipped | — |

`barspoon` now exits 1 in <1s with a clear message; unknown models still rejected;
summary/exit logic verified under `set -euo pipefail` with empty and populated arrays.

This also confirms the eval-path rewrite in #417 holds for **all** supported models
(#270/#274/#275 stay closed).

### DIFF
--- a/application/jobs/STAMP_classification/README.md
+++ b/application/jobs/STAMP_classification/README.md
@@ -72,7 +72,7 @@ All STAMP-specific variables use the `STAMP_` prefix to avoid collision with ODE
 
 | Variable | Default | Description |
 |----------|---------|-------------|
-| `STAMP_MODEL_NAME` | `vit` | Model architecture: `vit`, `mlp`, `trans_mil`, `linear`, `barspoon` |
+| `STAMP_MODEL_NAME` | `vit` | Model architecture: `vit`, `mlp`, `trans_mil`, `linear` (`barspoon` needs STAMP ≥ 2.5.0) |
 | `STAMP_FEATURE_TYPE` | *(auto-detect)* | Feature level: `tile`, `slide`, or `patient` |
 | `STAMP_DIM_INPUT` | `1024` | Input feature dimension (UNI/UNI2=1024, CTransPath=768) |
 | `STAMP_NUM_CLASSES` | `3` | Number of output classes |
@@ -118,7 +118,7 @@ Each site needs:
 | MLP | `mlp` | Multi-layer perceptron |
 | TransMIL | `trans_mil` | Transformer-based MIL |
 | Linear | `linear` | Linear classifier |
-| Barspoon | `barspoon` | Encoder-decoder transformer |
+| Barspoon | `barspoon` | Encoder-decoder transformer — **unavailable**: added in STAMP 2.5.0; image ships 2.4.0 |
 
 ## Quick Start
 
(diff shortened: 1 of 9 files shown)


Ultimate-Storm:
Two clean-ups on this branch:

**1. Dropped an unrelated commit.** This branch had accidentally picked up
`720fe42 "VPN: rewrite bare auth-user-pass…"`, which is not on `main`, so the diff
included `scripts/client_node_setup/setup_vpntunnel.sh`. Rebased to drop it — the
PR is now STAMP-only. That work is unaffected: it lives on
`fix/vpn-bare-auth-user-pass` (`f799667`), whose version of the file differs from the
dropped commit **only in comments** (identical `grep -qE` guards and
delete-then-append logic), so nothing is lost.

**2. The earlier `validate-swarm` failure is the known flake (#388), not this PR.**
The failing step is `Run dummy training in swarm` →
`scripts/ci/runIntegrationTests.sh run_dummy_training_in_swarm`, which drives the
**ODELIA** POC swarm (`admin@test.odelia`, polls the server log for
`Server runner finished.` with a 5-minute timeout). Evidence it cannot be caused by
this PR:

* #421 — which contains every change here except `run_stamp_deploy_test.sh` and two
  READMEs — **passed** that same step;
* `run_stamp_deploy_test.sh` is not invoked by `pr-test.yaml`;
* `setup_vpntunnel.sh` is not referenced anywhere in CI (and is now gone from the diff);
* #388 documents this exact step timing out under load.

The rebase re-triggered CI, so this run is also a re-test of that flake.

**Provenance note:** the DECADE kits already distributed to sites
(`jefftud/decade:1.5.0-dev.260710.a295c6b`) were built from the pre-rebase commit,
which is preserved as tag `decade-kit-1.5.0-dev.260710.a295c6b`.

PR #425, opened 2026-07-10
Evaluation: broaden metrics beyond Accuracy+AUROC (#410)

### LINKED ISSUE #410: Evaluation: broaden metrics beyond Accuracy+AUROC (sensitivity/specificity, F1, PR-AUC, calibration, per-class, CIs)
## Problem
The models compute **only Accuracy + AUROC** — for a clinical FL benchmark that is too thin to characterize model quality (class imbalance, operating point, calibration, and uncertainty are all invisible).

Current metrics ([application/jobs/_shared/custom/models/base_model.py:153-154](application/jobs/_shared/custom/models/base_model.py#L153)):
```python
self.auc_roc = ModuleDict({s: AUROC(...) for s in ["train_","val_","test_"]})
self.acc     = ModuleDict({s: Accuracy(...) for s in ["train_","val_","test_"]})
```
Logged as `{state}/ACC`, `{state}/AUC_ROC` in `_epoch_end` ([:174-178](application/jobs/_shared/custom/models/base_model.py#L174)). Nothing else is collected or reported.

## Fix — add a clinically-meaningful metric set
- [ ] Extend the torchmetrics `ModuleDict` in `base_model.py` with: **Sensitivity/Recall, Specificity, F1, Precision, PR-AUC (AveragePrecision/AUPRC), ConfusionMatrix, Calibration error (ECE)**, and **per-class** breakdowns (macro + per-class).
- [ ] Log them in `_epoch_end` under stable names (`{state}/F1`, `{state}/AUPRC`, `{state}/sensitivity`, …) so they're available to the ModelCheckpoint/selector and the monitor (#398).
- [ ] Surface the same metrics in the external eval path (`scripts/evaluation/predict.py` → `prediction_results.json`) so external validation reports the full set.
- [ ] Add **confidence intervals** (bootstrap over the eval set) in the predict/eval path for the headline metrics.
- [ ] Keep AUROC/ACC names unchanged (back-compat with `key_metric` and #409).

## Verification
- [ ] A short local training logs the new metrics each epoch.
- [ ] `predict.py` on a held-out set emits the extended metric set (+ CIs) in `prediction_results.json`.
- [ ] Multiclass and binary tasks both produce sensible values (e.g. AUPRC ~ prevalence baseline on random).

Ultimate-Storm:
Closes #410.

Adds a **tiered** metric set to `models/base_model.py`, attached to **val/test states only** so the per-training-step cost is unchanged:

| Tier | Metrics |
|---|---|
| always on | `F1`, `Recall` (sensitivity), `Specificity`, `PR_AUC` — macro |
| `ODELIA_EVAL_METRICS=full` | + per-class values + `ECE` (calibration) |
| `ODELIA_EVAL_BOOTSTRAP=1` | + `_ci_low`/`_ci_high`, **test split only** (the one expensive path) |

**Back-compat:** `{state}/ACC`, `{state}/AUC_ROC` and the explicit `val/AUC_ROC` log keys are preserved verbatim → `IntimeModelSelector` (`key_metric = val/AUC_ROC`) is unaffected.

Metric computation is factored into a **pure** `compute_epoch_metrics(state)` (no logging) so it is testable without a Trainer/GPU/dataset.

**Verification:** new `tests/unit_tests/test_base_model_metrics.py` (6 tests) — compat keys survive, macro scalars correct on separable inputs, train state keeps only the compat keys, `full` adds per-class+ECE, bootstrap adds CIs on test but **not** val. Ran the full unit suite locally on a CPU torch venv: **200 passed, 9 skipped, 0 failures**. Torchmetrics API assumptions verified empirically (logits accepted; `average=None` → per-class; `BootStrapper.compute()` → `{mean,std,quantile}`).

*Note:* CI resolves torchmetrics transitively via `lightning` (1.9.0 here) while the image pins `1.7.1` — all classes used exist in both; see the version-skew follow-up in #410.

### DIFF
--- a/application/jobs/_shared/custom/models/base_model.py
+++ b/application/jobs/_shared/custom/models/base_model.py
@@ -1,10 +1,42 @@
 from pathlib import Path
 import json
+import os
 import torch
 import torch.nn as nn
 import torch.nn.functional as F
 import pytorch_lightning as pl
-from torchmetrics import AUROC, Accuracy
+from torchmetrics import (
+    AUROC,
+    Accuracy,
+    AveragePrecision,
+    CalibrationError,
+    F1Score,
+    Recall,
+    Specificity,
+)
+
+# Extended evaluation metrics (#410) are attached to these states only, so the
+# per-training-step cost of a run is unchanged.
+EXTENDED_METRIC_STATES = ("val_", "test_")
+# Bootstrap CIs re-compute each metric N times over the buffered predictions, so
+# they stay on the test split even when enabled.
+BOOTSTRAP_METRIC_STATES = ("test_",)
+
+
+def _eval_tier():
+    """'default' = macro scalars; 'full' also adds per-class values + calibration."""
+    return os.environ.get("ODELIA_EVAL_METRICS", "default").strip().lower()
+
+
+def _bootstrap_enabled():
+    return os.environ.get("ODELIA_EVAL_BOOTSTRAP", "").strip().lower() in ("1", "true", "yes", "on")
+
+
+def _bootstrap_samples():
+    try:
+        return max(2, int(os.environ.get("ODELIA_EVAL_BOOTSTRAP_N", "200")))
+    except ValueError:
+        return 200
 
 
 class VeryBasicModel(pl.LightningModule):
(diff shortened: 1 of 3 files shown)


oleschwen (approved):
looks good

PR #426, opened 2026-07-10
Align the local best-checkpoint metric with the federated selector metric (#409)

### LINKED ISSUE #409: Local best-checkpoint monitors hard-coded val/ACC, not the model_selector key_metric (val/AUC_ROC) — local/federated best-model mismatch
## Problem

The **local** "best checkpoint" and the **federated** "best global model" are selected on **different metrics**, so for every AUROC-objective job the locally-saved `best_*.pt` is chosen on the *wrong* metric.

- **Local (per-client, per-epoch)** — Lightning `ModelCheckpoint` monitors a **hard-coded `val/ACC`** (accuracy), `mode="max"`:
  ```python
  # application/jobs/_shared/custom/threedcnn_ptl.py:533
  to_monitor = "val/ACC"
  min_max = "max"
  checkpointing = ModelCheckpoint(monitor=to_monitor, mode=min_max, save_top_k=1, save_last=True)
  ```
  `to_monitor` is a literal — it is never read from env/config/`key_metric` (only occurrence is line 533).

- **Federated (per-round, across the swarm)** — NVFlare `IntimeModelSelector` uses **`key_metric = "val/AUC_ROC"`** (all challenge jobs + `ODELIA_ternary_classification`; `config_fed_client.conf`).

So "best local epoch" (highest accuracy) and "best global round" (highest AUROC) optimize different objectives.

## Why this looks half-finished

`application/jobs/_shared/custom/models/base_model.py:174-178` logs **both** `val/ACC` and `val/AUC_ROC`, and explicitly re-logs AUROC with the comment **"For ModelCheckpoint, also log as 'val/AUC_ROC'"** — i.e. AUROC was exposed *for the checkpoint callback*, but `ModelCheckpoint.monitor` was never switched from `val/ACC` (or made `key_metric`-driven). This is the local counterpart of the earlier fix that set the federated selector's `key_metric` to `val/AUC_ROC`.

## Fix

Drive the local `ModelCheckpoint` from the **same `key_metric`** the federated selector uses, so local and federated agree:

- [ ] Plumb `key_metric` (from `config_fed_client.conf`'s `model_selector` args) to the training code — e.g. surface it as an env var alongside the existing `odelia_*` env vars (`env_config.py` / `env_vars`), or read it from config in the executor.
- [ ] In `threedcnn_ptl.py:533`, set `to_monitor = <key_metric>` (default `val/ACC` for back-compat) instead of the literal.
- [ ] Derive `mode` from the metric direction (max for `val/AUC_ROC` / `val/ACC`, min for `val/loss`) — mirror the `IntimeModelSelector` higher-is-better/`negate_key_metric` convention so the two never disagree.
- [ ] Confirm the chosen metric is logged under exactly that name at epoch end (`val/AUC_ROC` already is; `val/ACC`, `val/loss` too).
- [ ] Keep `save_last=True` so warm-continue's `mediswarm_latest_global.pt` mirror path is unaffected.

## Verification

- [ ] Unit/local: run a short local training with `key_metric=val/AUC_ROC`; confirm the retained `best_*.pt` corresponds to the epoch with the **highest AUROC**, not the highest accuracy.
- [ ] Confirm `best_FL_global_model.pt` / the challenge-eval path and the federated `IntimeModelSelector` now agree on the metric.
- [ ] Default (`key_metric` unset) still monitors `val/ACC` (no behavior change for jobs that want accuracy, e.g. `minimal_training_pytorch_cnn`).

## Refs
- `application/jobs/_shared/custom/threedcnn_ptl.py:533-542` (hard-coded `val/ACC`)
- `application/jobs/_shared/custom/models/base_model.py:174-178` (both metrics logged; the "for ModelCheckpoint" comment)
- `application/jobs/ODELIA_ternary_classification/app/config/config_fed_client.conf:162` (`key_metric = "val/AUC_ROC"`)

Ultimate-Storm:
Closes #409.

> **Re-framed after review — @oleschwen is right, and my original description overclaimed.** Correction below.

## What this is NOT
It is **not** a federated correctness bug, and "local best and global best come from different epochs" is **expected and harmless**. The model sent to the aggregator each round is the in-memory `pl_module.state_dict()` at the end of the round (`nvflare/app_opt/lightning/api.py:160-178`) — **no checkpoint file is ever read** by the aggregation path. `IntimeModelSelector` picks the global best out of the reported metrics dict on `val/AUC_ROC` (`intime_model_selector.py:112-126`), and that chain works: `best_FL_global_model.pt` was written in our one clean completed run (`34e07bb3`). Nothing about the local checkpoint touches it.

## What it actually is
The two arms select on **different metrics**:
- local Lightning `ModelCheckpoint` → hard-coded `monitor="val/ACC"` (`threedcnn_ptl.py:533`)
- federated `IntimeModelSelector` → `key_metric = "val/AUC_ROC"` (all 6 shared-training jobs)

In a swarm run the local checkpoint is inert. **But it is exactly the checkpoint the local-baseline evaluation loads** for held-out test scoring: `scripts/evaluation/run_duke_split_local_matrix.sh:554-574` resolves `best_checkpoint.json` → `epoch=N.ckpt` and hands it to the prediction step.

So in our reports, **local baselines are selected by ACC while federated results are selected by AUROC** — and on this imbalanced task the two diverge sharply (our clean run: ACC 0.789 vs AUROC 0.588). The local-vs-federated comparison is therefore not apples-to-apples.

**This PR aligns the two arms' selection criterion. It is a consistency/reporting fix with no effect on federated aggregation.**

## How
The training script is a `SubprocessLauncher` child and cannot read the NVFlare component config, so each job forwards its selector metric as a `KEY_METRIC=…` launcher prefix (the fork's launcher already strips leading `KEY=VALUE` tokens into the env, composing with the existing `MODEL_NAME=…`). `resolve_key_metric()` reads it, defaults to `val/AUC_ROC`, infers `mode=min` for loss-like metrics, and honors a `KEY_METRIC_MODE` override.

⚠️ **Behavior change (intended):** the local best checkpoint now tracks AUROC, not ACC — so local-baseline numbers in future reports will differ from past ones. `minimal_training` / `STAMP` / `cifar10` use different training code and are untouched.

## Verification
The value now lives in two places, so `tests/unit_tests/test_key_metric_consistency.py` asserts every shared-training job's launcher `KEY_METRIC` equals its selector `key_metric` (pure text check → always runs in CI). Confirmed it **fails** both when a selector drifts and when a launcher prefix is removed. Plus 5 `resolve_key_metric` unit tests. Full local suite: 206 passed, 9 skipped.

### DIFF
--- a/application/jobs/_shared/custom/threedcnn_ptl.py
+++ b/application/jobs/_shared/custom/threedcnn_ptl.py
@@ -36,6 +36,26 @@ FILENAME_GT_PREDPROB_SITE_MODEL_VALIDATION = 'site_model_gt_and_classprob_valida
 
 OMP_THREAD_ENV_VARS = ("OMP_NUM_THREADS", "MKL_NUM_THREADS", "OPENBLAS_NUM_THREADS", "NUMEXPR_NUM_THREADS")
 
+# Default matches the `key_metric` of every job that uses this training script
+# (application/jobs/*/app/config/config_fed_client.conf). Those jobs forward their
+# selector metric as a KEY_METRIC=... prefix on the launcher command, which the
+# SubprocessLauncher strips into the environment.
+DEFAULT_KEY_METRIC = "val/AUC_ROC"
+
+
+def resolve_key_metric():
+    """Return (metric, mode) for the local best-model checkpoint.
+
+    Keeps Lightning's ModelCheckpoint in step with NVFlare's IntimeModelSelector.
+    The training script runs in a separate subprocess and cannot read the NVFlare
+    component config, so the job forwards `key_metric` through the environment (#409).
+    """
+    metric = os.environ.get("KEY_METRIC", "").strip() or DEFAULT_KEY_METRIC
+    mode = os.environ.get("KEY_METRIC_MODE", "").strip().lower()
+    if mode not in ("min", "max"):
+        mode = "min" if "loss" in metric.lower() else "max"
+    return metric, mode
+
 
 @dataclass
 class UIDwithHash:
@@ -530,8 +550,8 @@ def prepare_training(logger, max_epochs: int, site_name: str = None,
 
         logger.info(f"Using model: {env_vars['model_name']}")
 
-        to_monitor = "val/ACC"
-        min_max = "max"
+        to_monitor, min_max = resolve_key_metric()
+        logger.info(f"Local best checkpoint monitors {to_monitor} (mode={min_max})")
         log_every_n_steps = 50
 
         checkpointing = ModelCheckpoint(
--- a/application/jobs/ODELIA_ternary_classification/app/config/config_fed_client.conf
+++ b/application/jobs/ODELIA_ternary_classification/app/config/config_fed_client.conf
@@ -83,7 +83,7 @@
       id = "launcher"
       path = "nvflare.app_common.launchers.subprocess_launcher.SubprocessLauncher"
       args {
-        script = "python3 custom/{app_script}  {app_config} "
+        script = "KEY_METRIC=val/AUC_ROC python3 custom/{app_script}  {app_config} "
         launch_once = true
       }
     }
--- a/application/jobs/challenge_1DivideAndConquer/app/config/config_fed_client.conf
+++ b/application/jobs/challenge_1DivideAndConquer/app/config/config_fed_client.conf
@@ -83,7 +83,7 @@
       id = "launcher"
       path = "nvflare.app_common.launchers.subprocess_launcher.SubprocessLauncher"
       args {
-        script = "MODEL_NAME=1DivideAndConquer python3 custom/{app_script}  {app_config} "
+        script = "KEY_METRIC=val/AUC_ROC MODEL_NAME=1DivideAndConquer python3 custom/{app_script}  {app_config} "
         launch_once = true
       }
     }
--- a/application/jobs/challenge_2BCN_AIM/app/config/config_fed_client.conf
+++ b/application/jobs/challenge_2BCN_AIM/app/config/config_fed_client.conf
@@ -83,7 +83,7 @@
       id = "launcher"
       path = "nvflare.app_common.launchers.subprocess_launcher.SubprocessLauncher"
       args {
-        script = "MODEL_NAME=2BCN_AIM python3 custom/{app_script}  {app_config} "
+        script = "KEY_METRIC=val/AUC_ROC MODEL_NAME=2BCN_AIM python3 custom/{app_script}  {app_config} "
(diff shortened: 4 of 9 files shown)


application/jobs/_shared/custom/threedcnn_ptl.py
```diff
@@ -36,6 +36,26 @@
...
+
+
+def resolve_key_metric():
+    """Return (metric, mode) for the local best-model checkpoint.
+
+    Keeps Lightning's ModelCheckpoint in step with NVFlare's IntimeModelSelector.
+    The training script runs in a separate subprocess and cannot read the NVFlare
+    component config, so the job forwards `key_metric` through the environment (#409).
+    """
+    metric = os.environ.get("KEY_METRIC", "").strip() or DEFAULT_KEY_METRIC
```
oleschwen, line 53: Couldn't this be simplified to `os.environ.get("KEY_METRIC", DEFAULT_KEY_METRIC)`?

Ultimate-Storm:
@oleschwen — you're right, and I've re-framed the PR (title + description).

> local best checkpoint and global best model from different epochs is probably fine and expected. You probably mean that the problem is different metrics used to select it, don't you?

Yes. I traced it to be sure, and my original framing was wrong on the important point:

- **Aggregation never reads a checkpoint.** `FLCallback.on_train_end` sends `pl_module.cpu().state_dict()` — the live module after `trainer.fit()` (`nvflare/app_opt/lightning/api.py:160-178`). The Lightning `ModelCheckpoint` is not in that path at all, so different epochs is exactly as harmless as you say.
- **Global-best selection is working.** `IntimeModelSelector` reads `val/AUC_ROC` out of the reported metrics dict (`intime_model_selector.py:112-126`) → `GLOBAL_BEST_MODEL_AVAILABLE` → `best_FL_global_model.pt`. I confirmed empirically: that file exists for our one clean completed run (`34e07bb3`) and is absent only from partner runs that never finished a round.

So the real issue is the narrower one you identified — **the differing metric** — and it does have a consequence worth fixing: the local `val/ACC` checkpoint is inert in the swarm, **but it is the checkpoint `scripts/evaluation/run_duke_split_local_matrix.sh:554-574` loads for the local-baseline held-out evaluation.** So our reports compare an ACC-selected local baseline against an AUROC-selected federated result. With ACC 0.789 vs AUROC 0.588 on this imbalanced task, that comparison isn't apples-to-apples.

Downgraded from "correctness bug" to "consistency/reporting fix" accordingly. The code is unchanged; only the claim is.

oleschwen (approved):
looks good

PR #423, opened 2026-07-09
Data guard: surface failure-reason breakdown (unreadable vs degenerate) so sites don't delete valid data (#416)

Ultimate-Storm:
Closes #416. The degenerate-input abort previously said *fix/exclude the data* for every failure type; when the real reason was `load_error:PermissionError` (unreadable files) it led a site to **delete valid data** instead of fixing permissions (CAM, 2026-07). Now the message carries a per-reason breakdown and the correct remedy: all-file-access -> *PERMISSION/OWNERSHIP problem, do NOT exclude*; degenerate -> *re-export/exclude*; mixed -> both counts; plus the per-UID jsonl path. Message builder is a pure staticmethod, validated standalone (all-perm / all-degenerate / mixed). Compiles clean.

Issue #372, opened 2026-07-01
ODELIA: Include evaluation on test data in local/swarm training


oleschwen:
In addition to evaluating model performance on possibly other datasets, it would be useful to evaluate the resulting model on the local test dataset at the end of local and swarm trainings, outputting prediction probabilities in the same format as for training and validation data during training.

oleschwen:
Implement as part of application code.

oleschwen:
The mechanism to control every how many epochs prediction probabilities shall be computed and stored as csv is not working as intended: it is only used for the aggregated models, the local models after each epoch during a round are always subject to inference (via the callback).

- [ ] decide whether mechanism should be removed (The intention was to reduce computational workload, but is this still a bottleneck? Saving predictions more often gives us more flexibility to evaluate metrics after training.) or implemented in the intended way.
- [ ] implement what we decided (renaming either the `should_export_aggregated_predictions` function or the `ODELIA_PREDICTION_EXPORT_EVERY_N_ROUNDS` env var if we keep it)

oleschwen:
This should be tested for all models (maybe already covered by weekly/manual test?)

PR #427, opened 2026-07-10
Enable k-fold cross-validation (#411)

### LINKED ISSUE #411: Enable k-fold cross-validation (fold infra exists but only fold 0 is used)
## Problem
Cross-validation is **not run** even though the data pipeline is fold-aware. The split (`split_df`) has a **`Fold`** column and the dataset can select `fold=N` ([dataset_3d_odelia.py:56-84](application/jobs/_shared/custom/data/datasets/dataset_3d_odelia.py#L56)), but **every call hard-codes `fold=0`** ([:206](application/jobs/_shared/custom/data/datasets/dataset_3d_odelia.py#L206); `load_split(..., fold=0)`) and there is **no fold loop** in the datamodule, `main.py`, or `threedcnn_ptl.py`. So results come from a single fixed train/val/test split — high variance on small clinical cohorts.

## Fix — make the fold selectable and support a CV sweep
- [ ] Replace the literal `fold=0` with a configurable **`FOLD`/`fold`** value plumbed from `env_config.py` (env var, default 0 for back-compat) through the datamodule/dataset.
- [ ] Add an optional **CV driver** that runs folds `0..K-1` and aggregates metrics (mean ± CI across folds). For **local training** this is a simple loop; for **swarm** it means one full run per fold.
- [ ] Document the **federated-CV caveat**: all sites must agree on identical fold definitions in their split files, and a k-fold sweep multiplies swarm-run wall-clock by K.
- [ ] Report the per-fold + aggregate metrics in the run artifacts.

## Verification
- [ ] `FOLD=1` trains/evaluates on fold 1's split (distinct UIDs from fold 0).
- [ ] A local k-fold sweep produces per-fold metrics + an aggregate mean/CI.
- [ ] Default (unset) still trains fold 0 — no behavior change for existing jobs.

Ultimate-Storm:
Closes #411.

`split.csv` already carries a `Fold` column (5-fold nested StratifiedGroupKFold, grouped by `PatientID`) and `ODELIA_Dataset3D` already accepts `fold=` — but every caller inherited the hard-coded `fold=0` default, so **only fold 0 was ever trained**.

**Design:** `fold` is a **run-level** parameter. NVFlare ships one job to every client, so all sites must train the same fold → K-fold CV is **K sequential swarm runs**, not a loop inside one run. `FOLD` defaults to `0` (today's behavior).

- `env_config.py`: `FOLD` env var (validated 0..4) threaded into all six `ODELIA_Dataset3D` constructions; `run_name` gains a `foldN` label so artifacts can't collide.
- `prepare_odelia_job.sh --fold N` → `patch_warm_start_job.py` patches the launcher (`FOLD=N`, which `SubprocessLauncher` strips into the child env, composing with `MODEL_NAME=`).

⚠️ **The real hazard, fixed:** `latest_global_path` was a *single fold-agnostic file*, so `--warm-start continue` on fold 1 would have resumed from **fold 0's global model** — cross-fold leakage. Fold N>0 now gets its own mirror (`…_foldN.pt`); fold 0 keeps the legacy filename for back-compat. `collect_swarm_globals.sh` already accepts `-p`, so it needs no change.

**Verification:** new `test_fold_selection.py` (9 tests: default/env/blank/out-of-range/non-integer, settings carry fold, every dataset built for the fold, run dir namespaced, no-aug path) + 8 new admin-patcher tests (launcher token idempotent + preserves other tokens; fold 0 legacy mirror; fold N own mirror; idempotent). Ran end-to-end against a **real** job config: `--fold 2` rewrites launcher + both mirror paths, `--fold 0` preserves legacy, `--fold 7` rejected. Full suite: **213 passed, 9 skipped, 0 failures**.


Ultimate-Storm:
@oleschwen — all three answered, and two of them turned out to be real gaps. Pushed as two commits.

### 1. "Should the number of folds be hard-coded or inferred from split.csv?"

**Neither — the fold shouldn't be validated against a *count* at all.** `NUM_FOLDS = 5` was duplicated in two files, and 5 is only what the split generators happen to use (`step4_create_split.py:12`, `create_synthetic_dataset.py:26`) — it is not an invariant. A site that regenerated with `n_splits=3`, or shipped a legacy `split.csv`, would sail past `fold < 5` and then fail deep in training.

So the **data is now the authority**: `env_config` checks only the type/sign, `NUM_FOLDS` is gone, and whether the fold *exists* is validated against that site's own `split.csv`. That is strictly stronger than any count — and because the check runs *after* the annotation merge and the image-presence filter, it also catches "the fold is in `split.csv` but this site has no images for it", which a pure fold-count check cannot see.

(The admin CLI keeps a 0–4 range as a **typo guard only** — `--fold 50` shouldn't reach six sites — and now says so in a comment. The admin cannot know each site's fold count.)

### 2. "A missing fold should be caught during preflight — but how? Fold empty, so client reports no data, and what happens then?"

Today: nothing good. I traced it.

- **Empty `train`** → the dataset builds fine with `len()==0`, survives the label-distribution logging, the class-weight block, and `compute_weighted_epochs` (which *explicitly tolerates* `num_train_samples <= 0`), and finally dies in PyTorch at `datamodule.py:75-76` with:
  `ValueError: num_samples should be a positive integer value, but got num_samples=0`
  — after minutes of UID scanning and hashing, naming neither the fold, nor the split, nor `split.csv`.
- **Empty `val`/`test`** → **worse: it raises nothing at all.** `shuffle=False` → `SequentialSampler` → zero batches. Lightning just skips validation, and `ModelCheckpoint` then monitors a metric that is never computed. Silent.
- In a swarm run, that client exits non-zero and simply doesn't contribute. With `min_responses_required=5` of 6, **one** such site is absorbed almost silently; **two** stall the round. Either way you burn a multi-hour run to discover a `split.csv` was wrong.

Now it fails **fast and legibly**, at the single choke point all six dataset constructions funnel through:

```
ODELIA: CAM_1 has no samples for FOLD=3 / split='train'.
split.csv contains folds [0, 1, 2] -- fold 3 is not among them.
Check /data/CAM_1/metadata_unilateral/split.csv.
```

`train`/`val`/`test` are all fatal (the empty-`val` case above is exactly why `val` must be), with `ODELIA_ALLOW_EMPTY_SPLITS` as an escape hatch for a site that genuinely holds no test data. `log_UID_discrepancies` also now raises on a missing fold *before* the expensive image crawl, so the site is told in seconds.

### 3. "How is this meant to be configured, via docker.sh in the startup kit?"

**Both — and the split matters:**

- **Swarm (`--start_client`): the job, not `docker.sh`.** `prepare_odelia_job.sh --fold N` patches a `FOLD=N` prefix onto the launcher, which `SubprocessLauncher` strips into the training subprocess. One artifact, one person. A per-site swarm flag would let six operators disagree, and a mismatch would be **silent and scientifically fatal** — models trained on overlapping partitions, cross-fold test leakage, no error anywhere.
- **Local modes: `docker.sh` — and this was missing.** `--local_training` / `--preflight_check` / `--dummy_training` bypass NVFlare entirely, so nothing patches anything and `FOLD` can only arrive via the container env. It wasn't passed, so **those modes could only ever train fold 0** — meaning a site had no way to check it *has* fold N. Now added (`--fold`, following the `--num_epochs` CLI>env>default pattern). The usage block states `--start_client` ignores it, which is guaranteed: the launcher copies `os.environ` *before* applying the job's `KEY=VALUE` prefixes, so the job always wins.

Which gives you the workflow your question was really pointing at:
```
./docker.sh --preflight_check --fold 3 --data_dir ... --scratch_dir ...
```
→ fails in seconds if that site doesn't have fold 3.

**Requires a kit re-issue** (it touches `master_template.yml`) — worth batching with #400, which needs one anyway.

### Also worth flagging
Threading `fold` into `log_UID_discrepancies` is a genuine correctness fix hiding in a one-line hunk: **the discrepancy scan previously always reported on fold 0**, regardless of the fold being trained.

**Tests:** 7 new against the real dataset+manifest (6 of the 7 fail without the guard), plus 6 asserting the generated `docker.sh` exposes `--fold` and is valid bash. Suite: 232 passed.

oleschwen:
Configuring the fold via the job description for swarm trainings offers flexibility, but has the danger that one fold is checked in the preflight check and a different fold is used in training. Do we want to accept that risk?
Otherwise we could 
* stick to hard-coding the fold as before
* hard-code a list of folds to be checked in preflight and only allow those folds to be configured and used for training (even with quick failure and meaningful error messages, we would want to run one epoch per fold to make sure the data will be fully usable, i.e., the preflight check would take 5 times as long)
* (maybe there is a better approach)

Ultimate-Storm:
Good catch — the skew is real. But two facts make it cheaper to close than the options suggest, and bound the damage even today.

**1. The expensive check is fold-independent, so "one epoch per fold" isn't needed.**
`split.csv` is long-format: `create_synthetic_dataset.py:111` writes one row per **(UID × fold)**, so *every* image appears in *every* fold — a fold only changes the train/val/test **partition**, not the image set. The costly part of preflight (does each image load / is it degenerate / permission errors — `_load_source_image` + the degenerate guard, all keyed per-UID) is therefore **identical across folds**. One epoch on any fold already exercises the same images fold 3 would. So checking all folds does **not** cost 5×.

**2. The only per-fold thing is the partition, and that check is ~free.** `ODELIA_Dataset3D._require_non_empty` (this PR) validates that the configured fold has non-empty train/val/test purely from row counts — milliseconds, no training.

**3. The skew is already bounded to a fast, clear failure, not a silent wrong result.** `_require_non_empty` runs during dataset construction **in the swarm run too**, not just in preflight. So if preflight checked fold 3 but the admin's job trains fold 2 and this site's fold 2 is empty/broken, the client aborts at **round 0 in seconds** with `"<site> has no samples for FOLD=2 / split='train'. split.csv contains folds […]"` — you lose seconds, not a multi-hour run, and you never get a wrong result.

### Recommendation
Take your option 2, but derive the fold list from the data and skip the per-fold epoch:

- In `--preflight_check`, run the **cheap** `_require_non_empty` structural check for **every fold present in `split.csv`** (from `manifest.split_df['Fold'].unique()`), not just the configured one. Milliseconds.
- Keep the **one** expensive epoch on a single fold — it covers image usability for all folds (fact 1).

Net: preflight becomes authoritative for *whatever* fold the admin later configures, at essentially no extra cost, and there's no hard-coded fold list to drift. It's strictly better than reverting to a hard-coded fold (which just moves the same "is this fold usable here?" question to run time with no answer).

I'm happy to add the all-folds preflight loop to this PR if you agree — it's a small change on top of the guard that's already here. (Reverting to a single hard-coded fold is the one option I'd argue against: it removes the CV capability to avoid a risk that a ~10-line preflight loop closes.)

Ultimate-Storm:
Decision: **we accept the risk** — keeping the per-run configurable fold as this PR has it, no all-folds preflight loop.

What makes it acceptable: the skew can't produce a *silent wrong result*. `_require_non_empty` runs during dataset construction in the swarm run itself, so if the configured fold is empty/unusable at a site, that client aborts at **round 0 in seconds** with a named message (`"<site> has no samples for FOLD=N / split='train'…"`), not after a long run and not silently. Worst case is a fast, clear restart — an acceptable trade for keeping CV configurable.

No code change needed for this. Leaving the all-folds preflight loop as a documented follow-up if we later want preflight to be authoritative for every fold (it's cheap — folds share the same image pool — but not required). Ready for re-review.

oleschwen:
Good point about the same images being contained in multiple folds. However, this approach only checks all data with preflight check for a single epoch if
* test data is read as part of the preflight check (which should be part of #372, but is not yet the case)
* we rely on the csvs being generated by the script and not being modified for single folds afterwards when debugging the dataset, e.g., due to implausibiity messages.

Ultimately, the purpose of the data access preflight check is to reveal dataset problems before swarm training to the extent possible (if a file is deleted or becomes inaccessible after the preflight check, that's a problem we won't be able to capture), while being able to finish in a reasonable amount of time, also for production-size datasets.

I would suggest going with the current implementation for now and creating a follow-up issue to investigate potential improvements.

Ultimate-Storm:
Conflicts resolved and pushed. Only one was real — `docker_config/master_template.yml`, where #427's `--fold` and #449's `--image` each add an arg-parser line + a usage line; I kept both. `dataset_3d_odelia.py` and `threedcnn_ptl.py` auto-merged cleanly: the fold guard, #423's failure-reason breakdown, and #409's `key_metric` all coexist. Full unit suite green — **302 passed, 8 skipped**.

Heads-up: the merge push dismissed your approval (`REVIEW_REQUIRED` again) — a quick re-approve and it's mergeable.

oleschwen (approved):
Code looks fine, but CI should pass before merge
(last CI failure looks like a sporadic problem to me)

PR #404, opened 2026-07-06
Backbone: document canonical NVFlare fork patch set

Ultimate-Storm:
Part of #399.

Adds `docs/NVFLARE_FORK_PATCHES.md` — the curated MediSwarm-on-NVFlare patch list, the net-zero churn to drop when curating a rebase branch, the `START_TASK_TIMEOUT=10` review item, and upstreaming candidates. De-risks the 2.8.0 rebase (#392 P1).

The git-history rewrite on the fork + the timeout change live in the submodule and need a rebuild, so they are deferred to the build window.

PR #436, opened 2026-07-11
CI: stop failing integration tests on any 'error' substring (#434)

### LINKED ISSUE #434: CI: integration tests fail on any 'error' substring — NVFlare's shutdown audit race makes POC/simulation mode flaky
## Problem

The simulation- and POC-mode integration tests assert success with a blanket substring scan:

```bash
if grep -qi "Epoch 9: 100%" <<< "$OUTPUT" && ! grep -qi "error" <<< "$OUTPUT"; then
```
(`scripts/ci/runIntegrationTests.sh:214` and `:229`)

**Any** line containing the substring `error` fails the build — including log noise that has nothing to do with training.

NVFlare's audit logger races with shutdown and intermittently emits:

```
root - ERROR - Traceback (most recent call last):
  File ".../nvflare/fuel/hci/server/audit.py", line 40, in pre_command
  File ".../nvflare/fuel/sec/audit.py", line 61, in add_event
    self.audit_file.write(line + "\n")
ValueError: I/O operation on closed file.
```

This happens **after** training has completed successfully, yet turns the run red.

### Evidence it is flaky, not a real failure
| PR | run | `I/O operation on closed file` | POC result |
|---|---|---|---|
| #404 | passed | 0 occurrences | `✅ succeeded` |
| #400 | failed | 1 occurrence | `❌ failed` |

#400's entire diff is `docker run --restart=on-failure:5` → `--restart=unless-stopped` in `docker_config/master_template.yml` plus a doc line. The POC path runs `_run_test_in_docker` (`docker run --rm <image> <script>`, `:93-99`) and never uses the startup-kit `docker.sh`, so that diff **cannot** affect it. The two runs differ only by timing.

Impact: a green/red signal that depends on a shutdown race. Contributors are told their PR broke training when it did not — and the natural response (rerun until green) erodes trust in CI.

## Fix
- [ ] Replace the blanket `! grep -qi "error"` with a check that ignores this known-benign shutdown traceback, via one shared helper used by both call sites.
- [ ] Keep the positive assertion (`Epoch 9: 100%`) as the real success signal.
- [ ] Leave every other `error` line fatal — this must narrow the filter, not disable it.

## Verification
- [ ] Sample output containing only the audit-shutdown traceback → treated as success.
- [ ] Sample output containing a genuine `ERROR` (e.g. `CUDA error`, `RuntimeError`) → still fails.
- [ ] Sample output missing `Epoch 9: 100%` → still fails.

Ultimate-Storm:
Closes #434.

The simulation- and POC-mode tests asserted success with `! grep -qi "error" <<< "$OUTPUT"`, so **any** line containing "error" failed the build. NVFlare's audit logger races with shutdown and intermittently emits a harmless traceback *after* training has already finished:

```
root - ERROR - Traceback (most recent call last):
  File ".../nvflare/fuel/sec/audit.py", line 61, in add_event
    self.audit_file.write(line + "\n")
ValueError: I/O operation on closed file.
```

so the same commit passed or failed purely on **timing**.

**This is why #400 was red.** Its failing log contains that traceback once; #404's passing log contains it zero times. #400's diff only changes `docker run --restart=` flags in the kit template, which the POC path (a plain `docker run --rm`) never even uses.

Extracts the check into `scripts/ci/_error_filter.sh` so it can be unit-tested, and **narrows** it — it does not disable it. Whitelist covers only that shutdown traceback; every other "error" line stays fatal, **including a real error appearing alongside the benign noise**.

**Verification:** `tests/unit_tests/test_ci_error_filter.py` (9 tests) using the verbatim traceback from #400's failing log. Confirmed the old assertion fails on it and the new one passes, while `RuntimeError`/`CUDA error`/`docker: Error response` all still fail.

### DIFF
--- a/scripts/ci/_error_filter.sh
+++ b/scripts/ci/_error_filter.sh
@@ -1,0 +1,25 @@
+#!/usr/bin/env bash
+# Shared error-detection helper for the integration tests.
+#
+# The tests used to assert success with a blanket `! grep -qi "error"`, so ANY
+# line containing "error" failed the build. NVFlare's audit logger races with
+# shutdown and intermittently emits a harmless traceback *after* training has
+# already finished:
+#
+#     root - ERROR - Traceback (most recent call last):
+#       File ".../nvflare/fuel/sec/audit.py", line 61, in add_event
+#         self.audit_file.write(line + "\n")
+#     ValueError: I/O operation on closed file.
+#
+# That made the same commit pass or fail purely on timing (#434). This narrows
+# the filter -- it does not disable it. Any other line containing "error" is
+# still fatal.
+
+_benign_error_noise() {
+    grep -viE "I/O operation on closed file|fuel/sec/audit\.py|hci/server/(reg|audit)\.py|self\.audit_file\.write|root - ERROR - Traceback"
+}
+
+# has_real_error <output>  -> exit 0 if a non-benign "error" line is present
+has_real_error() {
+    grep -i "error" <<< "$1" | _benign_error_noise | grep -q .
+}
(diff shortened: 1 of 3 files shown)


oleschwen (approved):
looks good – allowlisting specific irrelevant errors makes sense

Issue #275, opened 2026-04-08
STAMP: Validate federated model quality against local-only baseline


Ultimate-Storm:
## Context

The STAMP deploy test currently validates that swarm training **completes** successfully, but does not compare the federated model quality against a local-only training baseline.

In the initial deploy test run (2 rounds, VIT, synthetic data), both clients achieved AUROC 1.0 — but this is on synthetic data with easy-to-separate classes and is not a meaningful quality benchmark.

## Proposed Solution

1. **Automated local baseline**: Before swarm training, run \`--local_training\` on each client's data and record final metrics (val_loss, val_auroc)
2. **Comparison**: After swarm training completes, compare federated metrics against local baselines
3. **Reporting**: Output a comparison table in the deploy test results JSON:
   \`\`\`json
   {
     "federated_auroc": 0.85,
     "local_auroc_site_1": 0.78,
     "local_auroc_site_2": 0.72,
     "improvement": "+9.7%"
   }
   \`\`\`
4. Optionally fail the test if the federated model is significantly **worse** than the best local model (regression check)

## Why This Matters

The core value proposition of swarm learning is that the federated model should perform at least as well as (ideally better than) any single site's local model. Without this check, we can't validate that model aggregation is actually beneficial.

## Related Files

- \`scripts/deploy/run_stamp_deploy_test.sh\` — deploy test orchestrator
- \`docker_config/master_template_STAMP.yml\` — has \`--local_training\` flag
- \`application/jobs/STAMP_classification/app/custom/main.py\` — supports \`local_training\` mode

Ultimate-Storm:
## Implementation approach (planned — deferred for now)

Add a local-only baseline + comparison to `run_stamp_deploy_test.sh`.

1. **Before** swarm training, run `--local_training` on each client's data (the STAMP startup-kit `docker.sh` supports it) and capture final val metrics (`val_loss`, `val_auroc`/`val_cindex`) from `ValidationMetricCallback` (`stamp_training.py:237-254`).
2. **After** swarm completes (and #270's eval), compare federated vs per-site local metrics and emit a table into the results JSON (`RESULTS_DIR`, L168), e.g. `{federated_auroc, local_auroc_site_*, improvement}`.
3. Optional **regression gate**: fail the deploy test if the federated model is materially worse than the best local model.

[...]

PR #442, opened 2026-07-13
STAMP: local baseline trained 1 epoch; prediction callback left the model in eval mode

Ultimate-Storm:
Both found by **UM Mainz** from a real Step-5 run. Both are training-correctness
bugs, not cosmetics.

## 1. `--local_training` trained for a single epoch

`main.py` took the epoch budget from `NUM_EPOCHS` (default **1**) for *both*
`preflight_check` and `local_training`, and passed it to `prepare_training()` as
`max_epochs` — overriding `STAMP_MAX_EPOCHS` (STAMP's default: 32).

One epoch is correct for the preflight smoke test. But the **local baseline is what
the swarm model is compared against** (`--compare-local`, #275), so every site's
baseline was effectively untrained and the comparison meaningless.

`local_training` now defaults to `STAMP_MAX_EPOCHS`; `preflight_check` keeps 1;
`NUM_EPOCHS` still overrides both.

**Verified** (real run, synthetic site):

| mode | last epoch |
|---|---|
| `preflight_check` | `Epoch 0/0` (1 epoch) |
| `local_training` | `Epoch 31/31` (32 epochs) |

## 2. The prediction callback leaked eval mode into training

`STAMPPredictionCallback._write_predictions()` calls `model.eval()`, and nothing put
the module back. Lightning does not restore it either — so **every epoch after the
first, and every swarm round after the first** (NVFlare calls `fit()` once per
round), trained with **dropout disabled**. This is the
`Found N module(s) in eval mode at the start of training` warning sites reported.

Fixed by restoring the previous mode in a `finally` block.

**Verified** by instrumenting `on_train_batch_start` and measuring the module modes
in a real 3-epoch run:

| epoch | before | after |
|---|---|---|
| 0 | train (36/0) | train (36/0) |
| **1** | **eval (0/36)** | train (36/0) |
| **2** | **eval (0/36)** | train (36/0) |

### DIFF
--- a/application/jobs/STAMP_classification/app/custom/main.py
+++ b/application/jobs/STAMP_classification/app/custom/main.py
@@ -46,8 +46,19 @@ elif TRAINING_MODE in [TM_PREFLIGHT_CHECK, TM_LOCAL_TRAINING]:
     SITE_NAME = os.getenv("SITE_NAME")
     if not SITE_NAME:
         raise ValueError("SITE_NAME environment variable must be set for local training")
+    # The two modes want very different epoch budgets:
+    #   preflight_check  — a smoke test: 1 epoch is the point.
+    #   local_training   — the site's *baseline*, which the swarm model is compared
+    #                      against. It must train properly, so it follows
+    #                      STAMP_MAX_EPOCHS (STAMP's own default: 32).
+    # Defaulting both to NUM_EPOCHS=1 silently produced a 1-epoch baseline, making
+    # the swarm-vs-local comparison meaningless.
+    if TRAINING_MODE == TM_LOCAL_TRAINING:
+        default_epochs = os.getenv("STAMP_MAX_EPOCHS", "32")
+    else:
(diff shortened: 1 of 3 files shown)


Ultimate-Storm:
### Merged despite a red `validate-swarm`

The failure is **#388**, not this PR:

```
FATAL_SYSTEM_ERROR: client simulated_node_1 didn't report status for 300 seconds
❌ Minimal example simulation mode failed.
```

That step is the **ODELIA minimal-example** simulation/POC. Evidence it is environmental:

* a **docs-only** PR (#444, one markdown file) failed the *same* step today;
* this PR does not touch anything that step executes;
* the self-hosted runner shares a host with the DECADE builds/deploy tests, so the 300 s status timeout trips under load (exactly what #388 describes);
* an earlier commit on this branch **passed** the same job.

The change itself was validated end-to-end on real hardware (see the PR body).

PR #443, opened 2026-07-14
STAMP docker.sh: require a value for --data_dir/--scratch_dir/--GPU (empty var swallows the next flag)

Ultimate-Storm:
## Problem

A DECADE site ran the documented command with `SCRATCHDIR` never exported:

```console
$ ./docker.sh --scratch_dir $SCRATCHDIR --GPU device=0 --dummy_training
[INFO] No interactive terminal detected, disabling TTY.
Unknown parameter passed: device=0
```

The empty variable expands to nothing, so `--scratch_dir` silently consumes the
**next flag** (`--GPU`) as its value, leaving `device=0` orphaned. The error then
blames `--GPU` — which is perfectly valid usage — and the site went looking for a
problem with the GPU flag that does not exist.

## Fix

Validate that options which take a value actually receive one, and that the value
is not another flag. The message now names the real cause.

## Verification

| case | result |
|---|---|
| `SCRATCHDIR` unset (the site's case) | `❗ Option --scratch_dir requires a value, but got '--GPU'.` |
| `SCRATCHDIR` set | parses correctly |
| `--GPU` with no value at all | caught |
| genuinely unknown option | still rejected as before |

### DIFF
--- a/docker_config/master_template_STAMP.yml
+++ b/docker_config/master_template_STAMP.yml
@@ -714,12 +714,33 @@ docker_cln_sh: |
       TTY_OPT=""
   fi
 
+  # Require a real value for options that take one. An unset shell variable —
+  # e.g. `--scratch_dir $SCRATCHDIR` when SCRATCHDIR was never exported — expands
+  # to nothing, so the option silently swallows the *next* flag and the argument
+  # after that is reported as unknown. A site hit exactly this:
+  #     ./docker.sh --scratch_dir $SCRATCHDIR --GPU device=0 --dummy_training
+  #     Unknown parameter passed: device=0
+  # ...which blames --GPU, when the real problem is an empty $SCRATCHDIR.
+  _need_value() {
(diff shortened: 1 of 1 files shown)


Ultimate-Storm:
### Merged despite a red `validate-swarm`

The failure is **#388**, not this PR:

```
FATAL_SYSTEM_ERROR: client simulated_node_1 didn't report status for 300 seconds
❌ Minimal example simulation mode failed.
```

That step is the **ODELIA minimal-example** simulation/POC. Evidence it is environmental:

* a **docs-only** PR (#444, one markdown file) failed the *same* step today;
* this PR does not touch anything that step executes;
* the self-hosted runner shares a host with the DECADE builds/deploy tests, so the 300 s status timeout trips under load (exactly what #388 describes);
* an earlier commit on this branch **passed** the same job.

The change itself was validated end-to-end on real hardware (see the PR body).

Issue #416, opened 2026-07-09
Degenerate-input guard: surface the failure-reason breakdown (unreadable vs degenerate) — current wording led a site to delete valid data


Ultimate-Storm:
## Problem

The degenerate-input guard aborts with a message that **hides the actual failure reason**, and its wording actively pushes sites toward the wrong remedy. This has already caused a partner to **delete valid data**.

```python
# application/jobs/_shared/custom/data/datasets/dataset_3d_odelia.py:416-420
raise RuntimeError(
    f"ODELIA: too many degenerate/corrupt inputs "
    f"({self._bad_input_count} > {budget} = {self.MAX_BAD_INPUT_FRACTION:.0%} of {n}); "
    f"run the preflight scan (find_degenerate_inputs) and fix/exclude the data."
)
```

The per-input reason **is** captured — `_record_bad_input(...)` ([:323-333](application/jobs/_shared/custom/data/datasets/dataset_3d_odelia.py#L323)) writes `load_error:<ExceptionType>` ([:406](application/jobs/_shared/custom/data/datasets/dataset_3d_odelia.py#L406)) or a `DegenerateImageError.reason` ([:413](application/jobs/_shared/custom/data/datasets/dataset_3d_odelia.py#L413)) to `$SCRATCH/odelia_bad_inputs.jsonl` — but **none of it appears in the abort message**, and the message doesn't even name the file.

### Real incident (CAM, 2026-07)
All **579 records / 33 unique UIDs** in CAM's `odelia_bad_inputs.jsonl` had reason **`load_error:PermissionError`** — the container simply could not *read* those files. But the message said *"degenerate/corrupt inputs … fix/exclude the data"*, so the site (and the maintainer triaging it) concluded the images were corrupt and **removed 6 UIDs from `split.csv`**, discarding good data. Their preflight then "passed" only by dropping under the 5% tolerance (val 214 → 208); ~6 unreadable files remain **silently skipped**, and the 33-UID set can grow and re-abort a future run.

The two failure classes need **opposite** remedies:

| Reason | Meaning | Correct action |
|---|---|---|
| `load_error:PermissionError` / `OSError` | file unreadable by the container | **fix ownership/permissions** — do NOT exclude data |
| `DegenerateImageError` (all-zero/masked), decode failures | genuine bad image content | re-export or exclude |

## Fix

- [ ] Track reason counts (`collections.Counter`) alongside `_bad_input_count`.
[...]

PR #447, opened 2026-07-14
Fix the degenerate-input guard tests, and make CI actually run them (#416)

Ultimate-Storm:
**`main` is red right now** for anyone with `torchio` installed — and I broke it, in #423.

#423 added `self._bad_input_reasons` in `__init__` and writes to it from `_record_bad_input`, but the guard tests construct the dataset via `__new__` and never initialised it → 3 × `AttributeError`. A 4th test still asserted the pre-#423 abort wording.

### Why CI didn't catch it — the part that matters
`unit-tests.yaml` installs only `torch`, `lightning`, `scikit-learn`, `pytest`, `fastapi`. **No `torchio`, no `pandas`** — so `test_odelia_degenerate_input_guard.py` `importorskip`s out and is **silently skipped in every CI run**. That file covers the guard **CAM tripped over**, whose message once led a site to **delete six valid UIDs**. A green tick has meant nothing for it.

### This PR
- initialises `_bad_input_reasons` in the test fixture
- updates the abort assertion to the new reason-breakdown message
- **adds the regression that should have existed**: when every failure is a `PermissionError`, the message must say `PERMISSION/OWNERSHIP` and *"do NOT exclude the data"* — exactly the advice that was missing when CAM deleted valid data
- **installs `torchio` + `pandas` in CI**, so these tests (and #427's fold-validation tests) actually run

Suite with those deps present: **231 passed**. Worth merging promptly — `main` is broken.

### DIFF
--- a/.github/workflows/unit-tests.yaml
+++ b/.github/workflows/unit-tests.yaml
@@ -47,6 +47,10 @@ jobs:
           pip install lightning scikit-learn pytest
           # fastapi: server_tools/app.py imports it; needed by test_swarm_progress.py (#398)
           pip install fastapi
+          # torchio + pandas gate the ODELIA dataset tests (degenerate-input guard,
+          # fold validation). Without them pytest silently SKIPS those files, which is
+          # how a broken guard test reached main (#416/#423).
+          pip install torchio pandas
 
       - name: Run unit tests
         run: |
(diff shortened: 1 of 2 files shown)


oleschwen (approved):
looks good, we should wait for the CI to pass

PR #449, opened 2026-07-14
Kits: stop regenerating the PKI every build; encrypt kits; one consortium hub

Ultimate-Storm:
Partners get a new startup kit almost weekly. **That was a bug on our side, not a process problem.**

## Root cause (1 line)
The build version was embedded in the NVFlare **project name**. NVFlare keys its provisioning state off that name, and CertBuilder reuses the root CA + per-site keys only if that state dir already exists — so every build minted a **new root CA and a new client.key for all 8 sites**, silently invalidating every deployed kit. `workspace/` holds **49 ODELIA + 12 DECADE** provisioning runs.

Two builds 8 days apart:
```
odelia_1.5.0-dev.260702.be9ef04   rootCA md5=160ef17ba822
odelia_1.5.0-dev.260710.a6b0291   rootCA md5=9ce931deedb5   <- different trust root
```
The image tag was a red herring: the backbone, training code and job configs all live **in the image**, and docker.sh pulls by tag **every run**.

## After — same project, two different versions
| Build | rootCA | UKA client.key | DOCKER_IMAGE |
|---|---|---|---|
| testA | 1fd9c43ca2 | 27fcd53041 | ...:9.9.9-testA |
| testB | 1fd9c43ca2 OK | 27fcd53041 OK | ...:9.9.9-testB OK |

**A kit now survives an image bump.** Re-issue only when a site is added/removed or certs expire. (One final re-issue hands sites their stable-cert kits.)

## Also
- **--image / startup/image.conf** — move a site to a new release with no new kit. So *how do we ship frequent NVFlare/algorithm updates?* → **kits are not involved.**
- **Version-skew detection** — the guardrail. `expected_version` existed but was **never compared**; a site quietly on stale code was invisible. The heartbeat now reports the image *actually running*, and the odd site out gets a red chip.
- **Kits encrypted** (AES-256/PBKDF2-600k). They contain client.key and were plain zips. Deliberately **not** password-protected zip — that is ZipCrypto, which is broken and would not protect a private key on a shared link. Verified: no plaintext zip survives, key absent from ciphertext, wrong password rejected, sha256 matches manifest.
- **docs/consortium/** — handbook + 4 Sheet CSVs replacing the weekly email. Checklist rows are the *real* preflight checks; the 8 sites come from the provisioning yml. A test asserts every flag/script the handbook names actually exists.

Flagged, not hidden: the **live monitor has no auth or per-site scoping** (must not be shown to partners), and **STAMP kits have no pre-run checks** although three docs claim they do.

Suite: **254 passed** (3 failures are the pre-existing main breakage that #447 fixes).


oleschwen:
So far, I considered fresh certificates and a fresh CA in each startup kit and a 1:1 match of startup kit and image a feature: it prevents potentially incompatible startup kits from connecting rather than possibly producing silent problems. Do we want to drop that?
We certainly get more flexibility by allowing different versions of startup kits to work together and we can still force new CA, but also create potential for problems.

Ultimate-Storm:
> So far, I considered fresh certificates and a fresh CA in each startup kit and a 1:1 match of startup kit and image a feature: it prevents potentially incompatible startup kits from connecting rather than possibly producing silent problems. Do we want to drop that? We certainly get more flexibility by allowing different versions of startup kits to work together and we can still force new CA, but also create potential for problems.

Yes I wish to drop it, a feature preventing a project to move forward is worse than quality control/tracking

oleschwen:
Fine with me if we are aware that this can cause problems. Maybe we'll never encounter them.

scripts/build/_generateStartupKitArchives.sh
```diff
@@ -1,13 +1,77 @@
...
+    esac
+    [ -d "$startupkit" ] || continue
+
+    archive="${startupkit}_${LONG_VERSION}.zip"
+    encrypted="${archive}.enc"
+
+    zip -rq "$archive" "$startupkit"
+
+    password="$(openssl rand -base64 24)"
+    openssl enc -aes-256-cbc -pbkdf2 -iter "$PBKDF2_ITER" -salt \
```
oleschwen, line 62: This seems overkill for our current purposes and compared encrypted .zip files we previously obtained from the browser-based startup kit generation. On the other hand, if we introduce encryption, we might as well do it properly.

oleschwen (approved):
looks good, but we should wait for CI to pass.

Ultimate-Storm:
@oleschwen — you asked whether we want to drop the "fresh CA per build / 1:1 kit↔image match," since it was a deliberate fail-closed safety feature. Short answer: **we don't drop it — we keep it as a deliberate switch, and stop paying it as an every-build tax.** We validated all of this on the dl servers; evidence below.

## The trade, stated plainly
- The old behavior gave **fail-closed safety**: an incompatible kit couldn't connect, because a rebuilt server had a new CA.
- It also gave us **weekly kit churn**: *every* image build minted a new CA + new per-site keys, so every routine image bump invalidated every deployed kit — even when nothing about compatibility had changed. That safety was firing on every build, not just on real breaking changes.

## Recommendation: stable by default, fresh-CA on purpose, skew as the backstop

**1. Stable certs by default** (the merged #449). A kit survives a routine image bump.
*Validated:* two independent builds (different tags) produced a **byte-identical root CA** (30 files → one hash) and **byte-identical `client.key` for all 8 sites**; and an **image-A kit completed the TLS handshake and registered on an image-B server** (`New client RUMC_1@… joined`). So a routine bump no longer breaks kits.

**2. Your safety feature is retained — as an opt-in, not an every-build side effect.** When the CA *must* rotate (a real breaking protocol change, or key compromise), the operator forces a fresh PKI by provisioning under a **version-carrying project name** — exactly what the current DUKE test project still does. Old kits then fail closed, as before.
*Validated:* in the same run, provisioning `odelia_allsites` (de-versioned) reused the CA, while provisioning the version-carrying DUKE project minted a **fresh** CA. Both behaviors exist side by side today; which one you get is now a deliberate choice (the project name), not an accident of the build.

**3. Version-skew detection as the always-on backstop.** Because image and kit are now decoupled, a site could silently run divergent code. The monitor now flags that: the heartbeat reports the container's actual image, and the dashboard shows a red chip on any site off the swarm majority (or off the roster).
*Validated:* the odd-image-out site is flagged, the majority stays clean, a single site never false-positives. **Caveat you should weigh:** the chip is **advisory** — it warns, it does **not** abort the run. So for a genuinely breaking change, skew detection alone is *softer* than your fail-closed CA. That's the real decision on the table: for breaking changes, use the fresh-CA switch (2); rely on skew (3) only for catching accidental drift within a compatible line.

*(Also — the validation caught two bugs now fixed in #455: `docker.sh` let `image.conf` clobber an env override, and the skew code would 500 the dashboard the instant two sites disagreed. And the deploy orchestrators still expect plaintext `*.zip` — #456. Worth landing before we lean on any of this.)*

## Net
We're not choosing flexibility *over* safety. We're moving the fail-closed CA from "every build, whether or not it matters" to "the specific moments it matters," and adding skew detection to cover the in-between. If you'd rather skew be an **enforcement gate** (block a run when sites disagree) rather than advisory, that's a small, reasonable follow-up — say the word and I'll spec it.

PR #488, opened 2026-07-28
487 integration tests: remove swarm containers after killing them

### LINKED ISSUE #487: Remove Docker containers from locally run integration test
Due to the container restarting mechanism, Docker containers are not removed when running an integration test locally, preventing another run if they are not `docker rm`ed manually. (Not a problem in CI, there is a later step to remove them.)

oleschwen:
as the title says: remove swarm containers after killing them in integration test

### DIFF
--- a/scripts/ci/runIntegrationTests.sh
+++ b/scripts/ci/runIntegrationTests.sh
@@ -832,6 +832,7 @@ run_dummy_training_in_swarm () {
 kill_server_and_clients () {
     echo "[Cleanup] Kill server and client Docker containers if running ..."
     docker kill odelia_swarm_server_flserver_$CONTAINER_VERSION_SUFFIX odelia_swarm_client_client_A_$CONTAINER_VERSION_SUFFIX odelia_swarm_client_client_B_$CONTAINER_VERSION_SUFFIX || true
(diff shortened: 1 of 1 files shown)


Ultimate-Storm (approved):
nice, approved

Ultimate-Storm (approved):
LGTM — approving.

I checked the one thing that worried me: both the client and server containers are launched with `--restart=unless-stopped` (`docker_config/master_template.yml:1047` and `:1155`), so I wanted to be sure `docker kill` wouldn't let the restart policy revive the container and make the plain `docker rm` fail. Tested it empirically:

```
docker run -d --restart=unless-stopped --name probe alpine sh -c 'while true; do sleep 1; done'
docker kill probe   # -> State.Status = "exited"  (policy does NOT revive it)
docker rm probe     # -> succeeds
```

Docker treats an explicit `kill` via the API as a manual stop, so `unless-stopped` doesn't restart it. Your `kill` + `rm` sequence is correct as written.

Optional nit, take it or leave it: `docker rm -f` would collapse the two lines into one and also cover the case where the `kill` didn't apply (e.g. the container was already restarting). Not required — the `|| true` guards already keep this from failing the run.

Issue #480, opened 2026-07-27
Swarm: finalize_training after the while flare.is_running() loop never runs (launcher SIGTERMs the subprocess)


Ultimate-Storm:
## Summary

In swarm mode, `finalize_training()` after the `while flare.is_running()` loop **does not run**. Ole
noticed the post-loop code never executes; this confirms the mechanism and proposes a fix.

`application/jobs/challenge_1DivideAndConquer/app/custom/main.py`:

```python
while flare.is_running():          # main.py:59  — services every round
    ...
    threedcnn_ptl.validate_and_train(...)
# main.py:78-79 — intended to run once at the end, but is SIGTERM'd first:
if TRAINING_MODE in [TM_LOCAL_TRAINING, TM_SWARM]:
    threedcnn_ptl.finalize_training(...)
```

## Cause

`main.py` is launched **once per job** by `SubprocessLauncher` (`config_fed_client.conf`:
`launch_once = true`). On job end, `client_runner.end_run_events_sequence` fires `ABOUT_TO_END_RUN`
(TaskExchanger `notify_end` + closes the pipe → the subprocess's `flare.is_running()` will return
False), then almost immediately `END_RUN` → `LauncherExecutor.finalize` →
`SubprocessLauncher._stop_external_process`, which with **`shutdown_timeout` defaulting to `0.0`**
(not set in the job config) does `self._process.wait(0.0)` (returns instantly) then
`self._process.terminate()` (SIGTERM). The subprocess, blocked in the `is_running()` poll (~0.5 s)
between rounds, is killed before it can exit the loop and reach line 79. `main.py` installs no
SIGTERM handler, so it just dies. Net: **Ole is right** — the post-loop code does not reliably run.

## Impact — benign, but worth fixing

`finalize_training` (`threedcnn_ptl.py:655-678`) only writes `best_checkpoint.json` (a pointer) and
copies `last.ckpt → last_global_model.ckpt`, plus the "Training completed successfully." log line.
The real artifacts are unaffected:

- per-round `epoch=*.ckpt` / `last.ckpt` are written by Lightning's `ModelCheckpoint` **during** fit;
- the aggregated global is saved by the persistor and mirrored to `/scratch/mediswarm_latest_global.pt`
  (warm-continue) each round;
- eval already falls back to `epoch=*.ckpt` / `last.ckpt`.

So no data loss. What's lost is the best-pointer + `last_global_model.ckpt` convenience copy + the
[...]

PR #491, opened 2026-07-28
490 fixed runintegrationtest.sh: run_3dcnn_training_in_swarm

### LINKED ISSUE #490: ODELIA runIntegrationTest.sh run_3dcnn_training_in_swarm fails
With a default configuration of `min_clients = 5`, `runIntegrationTest.sh run_3dcnn_training_in_swarm` fails, as it only provides two clients. This test is not executed as part of the CI because it would take excessively long, so we didn't notice it before.

- [x] try running fewer epochs in run_3dcnn_training_in_swarm and run_3dcnn_local_training as part of successful CI
- [ ] decide if we want to keep it in
- [x] make test pass if run locally

See also #352

oleschwen:
Functional change and bug fix: modify swarm job description for test so that it passes with two clients.

Additional change: changed 3DCNN local and swarm training in test to run for fewer epochs than in production and included it in CI. I would suggest to keep this included for now (adds about 3:30 minutes, see https://github.com/KatherLab/MediSwarm/actions/runs/30364208515/job/90291149872?pr=491) for thoroughly testing #480 and #372, and re-consider the number of epochs and shifting it to a daily/weekly/similar test as part of #352.


Ultimate-Storm (approved):
good to have, approved

PR #495, opened 2026-07-28
fix(build): keep pretrained-weight layers above the version-stamped template (docker cache)

Ultimate-Storm:
Answers @oleschwen's question about `COPY master_template.yml` always being a cache miss, and fixes the expensive consequence.

## Why it's a cache miss (it is **not** mtime)
You were right to be suspicious of the `git archive` mtime theory — but that's not it. Docker does exclude mtime from `COPY` checksums, exactly as the docs say. The file's **content genuinely changes on every commit**.

`scripts/build/buildDockerImageAndStartupKits.sh` rewrites the template's placeholders *before* the build:

```
__REPLACED_BY_CURRENT_VERSION_NUMBER_WHEN_BUILDING_DOCKER_IMAGE__      -> <version>-dev.<date>.<git short hash>
__REPLACED_BY_CONTAINER_VERSION_IDENTIFIER_WHEN_BUILDING_DOCKER_IMAGE__ -> <git short hash>
```

Both embed the **commit hash** (and the version string also embeds the date), so `master_template.yml` in the build context differs on every commit. The COPY checksum legitimately changes → cache miss → and everything below it rebuilds: the `nvflare` pip install, then the multi-GB `torch_home_cache` / `hf_home_cache` layers.

## Fix
Since that layer can *never* be cached across commits, move the weight COPYs **above** it — which is exactly the workaround you proposed. A code-only commit now rebuilds only:

`COPY master_template.yml` → `pip install nvflare` → `COPY ./MediSwarm`

and the heavy weight layers stay cached. I documented the root cause inline so the ordering constraint isn't silently undone later.

## Possible follow-up (not in this PR)
If we also want the `nvflare` install to stay cached, the version stamping could move *after* the install (sed the template in site-packages via a build `ARG`), or happen at provision time instead of image-build time. That would make `master_template.yml` content-stable across commits. Happy to file that separately if you think it's worth it.

Note `Dockerfile_STAMP` has the same COPY of a stamped `master_template_STAMP.yml`, but no weight-cache layers, so the impact there is much smaller.

### DIFF
--- a/docker_config/Dockerfile_ODELIA
+++ b/docker_config/Dockerfile_ODELIA
@@ -15,8 +15,10 @@ ENV PYTHON_VERSION=3.10.14
 # 1. pip installs FIRST (most expensive to rebuild, change least often)
 # 2. apt installs AFTER (change more often due to CVE version bumps, but
 #    are fast to rebuild)
-# 3. NVFlare from source (changes with NVFlare updates)
-# 4. COPY pretrained weights (rarely changes)
+# 3. COPY pretrained weights (rarely change, but multi-GB layers)
+# 4. NVFlare from source + master_template.yml (the template is stamped with
+#    this build's version/commit hash, so it is a cache miss on EVERY commit --
+#    nothing cacheable may sit below it; see the CACHE NOTE at that stage)
 # 5. COPY source code (changes every build — must be last)
 #
 # The base pytorch image already includes python3 + pip, so pip installs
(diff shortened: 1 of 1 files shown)


oleschwen:
Good point that the file changes in every commit; as the person who implemented that string replacement, I should have known :-|
I'd say that this is a major improvement in cache usage and image layer storage and would address the nvflare cache miss only if this turns out to be a problem.

oleschwen (approved):
looks good

PR #502, opened 2026-07-29
480 ODELIA call finalize_training also in swarm training

oleschwen:
Implemented call to finalize_training via callback so that the code is actually executed after training also in swarm training mode (which was, so far, implemented and intended to run, but was not executed in our swarm training setup).

Additionally, 
* slightly extended test for output of 3DCNN local training (for consistency)
* consistently print captured output/files present iff something was missing

Contains #523 (which was cherry-picked from here).

Note: "finalize" here is only about saving last and best local models, not including running inference on test data.


Ultimate-Storm (approved):
Approved

**Two non-blocking nits**

- *Naming/coupling:* `GT_PredProb_Output_Callback` now also performs checkpoint finalization and carries `logger`/`model`/`checkpointing`/`path_run_dir`/`env_vars` for that purpose. A future reader looking for "where does finalize happen" is unlikely to look in a class named for CSV export. A separate small `FinalizeCheckpointsCallback` would keep the responsibility obvious — happy for that to be a follow-up, not a blocker.
- *Per-round cost:* `finalize_training` now runs every round rather than once. It's idempotent and mostly copies already-written checkpoints, but it does a `shutil.copy` of `last.ckpt` each round. Fine at current model sizes; worth remembering if checkpoints grow.

**Context:** I'd briefly opened a competing in-loop fix in #505 and reverted it once I saw this PR — #505 is now deploy-test-only, so there's no overlap.

oleschwen:
Good points, let me think about this some more before merging.
The name is indeed misleading, I was thinking ahead to #372 where finalize_training should also output PredProb (for the test data), but it does more, that's easy to fix or split in separate callbacks.
Calling finalize_training one per _round_ is something I specifically wanted to avoid (also in the interest of not running inference on the test data multiple times). Apparently the callback is not the right mechanism for this purpose, and a test running only one round does not capture that it is called multiple times. See https://github.com/KatherLab/MediSwarm/issues/480#issue-4986326421.

scripts/ci/runIntegrationTests.sh
```diff
@@ -895,7 +930,26 @@ run_3dcnn_local_training () {
         fi
     done
 
+    # The following fails; output is contained twice in the captured output (for unknown reasons) even though it is printed only once during training.
```
oleschwen: Actually, a number of lines are repeated at the end of the captured output. So far, I couldn't figure out why.

Issue #466, opened 2026-07-20
Backbone: replace the blanket 100h timeouts with sane per-operation timeouts + FT retry


Ultimate-Storm:
## Problem

The fork carries `d81d51d0d` — "set timeouts to 100 hours (360000s) for slow VPN connections" — applied broadly across `ccwf/common.py`, `client/flare_agent.py`, `f3/streaming/byte_{receiver,streamer}.py`, `utils/pipe/pipe_handler.py`, `server/admin.py`, `server/server_engine.py`.

It was the right emergency fix (short timeouts were killing slow-VPN sites), but it converts **every** dropped leg into a **permanent silent hang** instead of a recoverable error. Nothing times out, nothing retries, nothing logs — the run just stops advancing and looks alive.

## Evidence (2.8.0 canary, 2026-07-16/17)

Three runs on the cosmos↔dl0/dl2 loop, at **20, 6 and 2** rounds, all stalled identically:

- Observed mechanism: the aggregation peer finishes a 688 MB model download (`download tx ... done: status=finished elapsed=778s`), the round advances, and then a **one-shot "come download" control message is lost**. The sender waits forever; the receiver never arms its `NVFLARE_TENSOR_MIN_DOWNLOAD_TIMEOUT` (1800s) because from its side no transfer ever started.
- The link is tailscale **DERP-relayed** (`relay "fra"`, direct hole-punch blocked by the university firewall) and drops a leg roughly every 1–1.5 h — so any run longer than that window hits it.
- Server log then shows no activity for hours with the job still "running"; GPUs idle; no error anywhere.

This is **version-independent** — 2.7.2 behaved the same way, and it is *not* an NVFlare 2.8.0 regression (2.8.0 is otherwise validated end-to-end). It is very likely the same failure class behind partner sites reporting "0 rounds" on long multi-epoch rounds.

## Proposed fix

1. **Size timeouts to the operation, not to the worst case.** A per-chunk/inactivity timeout (2.8.0 already has `MIN_DOWNLOAD_TIMEOUT`, default 300s) is the right shape; a 100 h *task* timeout is not. Keep generous *end-to-end* budgets, but every wait should have a bound that is minutes-to-hours, not days.
2. **Retry the control leg.** The lost "come download" message is a single unacknowledged request — make it retried/idempotent (or ride `ReliableMessage`) so one dropped packet doesn't strand a round.
[...]

PR #514, opened 2026-08-14
Make strict ODELIA swarm runs reliable and exact-client by default

### LINKED ISSUE #513: Strict swarm gather can aggregate before the final contribution is accepted
## Problem

In strict all-client mode, `Gatherer._do_gather()` publishes the final trainer's `reply_time` before `aggregator.accept()` and the contribution-accepted event hooks finish. The monitor calls `is_done()` without the gather lock, so it can observe all replies and start aggregation while the final contribution is still being accepted.

This can produce a terminal `FINISHED:COMPLETED` job that did not actually aggregate every required site.

## Evidence

The exact-8 one-round canary `a8c6a75b-0666-49fa-a832-ba3c2d15f5d2` reported eight accepted contributions and completed, but the aggregator logged:

- UKA acceptance began at `16:35:35.497`
- aggregation began at `16:35:35.543` and logged `aggregating 7 update(s)` at `.545`
- `Contribution from UKA_1 ACCEPTED` was logged only at `.687`

No client dropped and no error was emitted. Metadata alone therefore gave a false strict-success signal.

## Required behavior

When every participating client is required:

1. A response must count only after `aggregator.accept()` returns true and the accepted-contribution event hooks finish.
2. A rejected contribution, bad return code, or exception must remain uncounted and fail the strict round.
3. The learn-task timeout must never trigger partial aggregation.
4. Tolerant configurations may retain their existing minimum-response behavior.

## Proposed fix

Use a strict-only deferred commit: leave `trainer_status.reply_time` unset while acceptance and event handling run, then publish it only after successful completion. Add a deterministic concurrency regression test that blocks the final `accept()` and asserts `is_done()` remains false until it is released. Cover rejection and event-handler exception paths as well.

## Validation evidence

Replacement canary `1e4bface-91f8-460b-bcf0-2e14d60b5d72`, using the deferred-commit fix:

- exact named eight clients started and finished
- exactly eight contributions were accepted
- aggregator logged `aggregating 8 update(s) at round 0`
- final result reached the other seven sites and server received `ALL_DONE`
- no `ERROR`, `FATAL`, `invalid_request`, `MODEL_UNRECOGNIZED`, timeout, retry, or duplicate signal

## Acceptance criteria

- [ ] deterministic race regression test
- [ ] strict bad/rejected/exception results remain uncounted
- [ ] strict timeout cannot aggregate fewer than all required trainers
- [ ] tolerant-mode semantics remain unchanged
- [ ] exact-all-client canary logs the same count for accepted and aggregated updates

Ultimate-Storm:
## What changed

- Keep producer-side large-model result references alive for the full long-run budget in all six productive ODELIA jobs.
- Retry only missing/raw-timeout learn-task deliveries, retaining acknowledgements already received.
- Make repeated delivery of the same round idempotent and reject stale rounds.
- Treat missing result-submission permission replies as transient while preserving explicit terminal replies.
- In strict mode, reject bad/rejected contributions, prohibit timeout-driven partial aggregation, and count a contribution only after `aggregator.accept()` plus contribution hooks complete.
- Make the admin preparation path exact-client safe:
  - bare production preparation defaults to the current named eight ODELIA sites;
  - one `--strict-clients` value derives mandatory clients, participating clients, and all quorum values;
  - conflicting values fail before job files are written;
  - preparation happens in a temporary sibling and preserves the last valid staged job until validation succeeds;
  - generated admin kits carry the matching patcher with the helper.
- Keep reusable source-job quorum values generic for two-client and other validation workflows; supplying explicit `--min-*` values opts into that custom/test policy.

## Why

Two independent production defects were reproduced.

1. In strict job `6e1d08b9`, CAM's result reference had the default 1,800-second lifetime. It expired before MHA received serialized-submission permission at 1,881.55 seconds, so the first fetch returned `no ref found` and the round desynchronized. The producer TTL is controlled by `download_complete_timeout`, not tensor inactivity timeout.

2. Canary `a8c6a75b` reported eight accepted contributions and `FINISHED:COMPLETED`, but actually logged `aggregating 7 update(s)`. The monitor observed the final trainer's `reply_time` and began aggregation while `aggregator.accept()` was still processing UKA. The strict path now publishes that completion marker only after successful acceptance and event handling.

The production policy is exact all-site participation. Retries preserve all eight; terminal site failure fails loudly. No pruning or partial continuation is introduced by this PR.

## Impact

Future image builds contain the controller and six corrected job configurations. Future generated admin kits contain the strict-default helper and its matching patcher. Existing client identity/certificate kits do not need to be rebuilt merely to run a new image, but the new admin preparation behavior requires the updated helper/patcher (automatically included in newly generated admin kits).

## Validation

- Replacement one-round canary `1e4bface-91f8-460b-bcf0-2e14d60b5d72`: exact eight started/finished, exactly eight accepted, `aggregating 8 update(s)`, final broadcast to all peers, server `ALL_DONE`, no errors.
- Current full job `87c5bbee-6f8b-41cb-bd37-e9625cbecee2`: rounds 0–11 each aggregated exactly eight updates; round 12 remains healthy with all eight connected at the latest audit.
- `378 passed, 7 skipped, 1 deselected` in the complete non-GPU unit suite; only pre-existing metric warnings.
- Focused deterministic tests cover the final-accept race, rejection/event ordering, strict timeout, retry target retention, explicit-reply preservation, abort handling, and duplicate delivery.
- `bash -n` for both modified shell scripts.
- Python compilation and `git diff --check`.
- Exact `jefftud/odelia:1.6.0` constructor/config compatibility, including `download_complete_timeout`.
- End-to-end admin-helper staging smoke against the 1.6.0 image produced exact mandatory/participating client lists and 8/8/8 quorum values.

Closes #513.
Partially addresses #466.

### DIFF
--- a/application/jobs/ODELIA_ternary_classification/app/config/config_fed_client.conf
+++ b/application/jobs/ODELIA_ternary_classification/app/config/config_fed_client.conf
@@ -15,8 +15,12 @@
         args {
           launcher_id = "launcher"
           pipe_id = "pipe"
-          # time to wait for final result transfer from subprocess (30 min — large models over VPN)
-          last_result_transfer_timeout = 1800
+          # Let the client job wait for the subprocess result transfer for the
+          # full long-run budget.
+          last_result_transfer_timeout = 86400
+          # Keep producer-side tensor refs alive through a sequential multi-site
+          # submission queue, until the permitted peer fetches them.
+          download_complete_timeout = 86400
           # time to wait for subprocess to call flare.init() (10 min)
           external_pre_init_timeout = 600
           # time to wait for peer to read sent message (24 h - ride out transient VPN stalls; see #346)
@@ -44,8 +48,18 @@
           learn_task_abort_timeout = 300
           # time for task acknowledgment incl. model weight streaming (24 h - ride out transient VPN stalls)
           learn_task_ack_timeout = 86400
+          # Retry a missing/TIMEOUT learn-task acknowledgment every hour, while
+          # retaining ACKs from other sites and respecting learn_task_timeout.
+          learn_task_scatter_attempt_timeout = 3600
+          learn_task_scatter_retry_interval = 5
           # time for final result acknowledgment (24 h - ride out transient VPN stalls)
           final_result_ack_timeout = 86400
+
+          # Permission is a small idempotent control request. Allow relayed replies
+          # time to arrive and retry a missing reply without reserving another slot.
+          request_to_submit_result_msg_timeout = 60
+          request_to_submit_result_interval = 5
+          max_concurrent_submissions = 1
 
           # ids must map to corresponding components
           persistor_id = "persistor"
--- a/application/jobs/_shared/custom/fault_tolerant_ccwf.py
+++ b/application/jobs/_shared/custom/fault_tolerant_ccwf.py
@@ -14,26 +14,26 @@ configs) that let a run survive one (or a few) such failures:
   instead of ``system_panic``. (Silent/dropped clients and the configure /
   end-workflow counts are already tolerated by stock NVFlare when ``min_clients>0``.)
 * ``FaultTolerantSwarmClientController`` + ``FaultTolerantGatherer`` -- the
-  rotating aggregator tolerates a peer that submits a bad result: it is logged,
-  not counted toward ``min_responses_required``, and acknowledged so the failure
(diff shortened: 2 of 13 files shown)


oleschwen (approved):
Looks good, even though I cannot claim I understand all the changes.

PR #516, opened 2026-08-27
ODELIA: Workaround for local data directory under /data

oleschwen:
This PR does not double-mount the local data directory if located inside /data, which caused a mount failure so far.
(commit pulled out from #391)


Ultimate-Storm:
`[[ "$MY_DATA_DIR" =~ "/data" ]]` is a substring match, not a path test. `/mnt/database/x` and `/home/u/data_backup` both contain `/data`, so the mount is skipped and the container cannot see the data — the failure this is meant to fix.

```bash
case "$MY_DATA_DIR" in
    /data|/data/*) ;;
    *) DOCKER_MOUNTS+=" -v $MY_DATA_DIR:$MY_DATA_DIR:ro" ;;
esac
```

Ultimate-Storm (approved):
Approving as-is on the understanding the /data prefix test is corrected in a follow-up; tracking separately.

PR #517, opened 2026-08-27
ODELIA extend plotting of evaluation

oleschwen:
This PR extends the ovo AUROC plots by evaluation on test data.
Only evaluation scripts are affected, no swarm learning infrastructure or application code.
(pulled out of #391)

### DIFF
--- a/scripts/evaluation/compute_ovo_auroc_for_test_evaluation.py
+++ b/scripts/evaluation/compute_ovo_auroc_for_test_evaluation.py
@@ -1,0 +1,59 @@
+#!/usr/bin/env python3
+
+import numpy as np
+import pandas as pd
+import sys
+from itertools import product
+from tqdm import tqdm
+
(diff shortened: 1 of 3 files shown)


Ultimate-Storm:
`predict.py`: `round(..., 16)` is right — 6 dp created ties that skew rank-based AUROC. It also drops a blank line before `print_metrics`; needs two between top-level defs (E302).

`compute_ovo_auroc_for_test_evaluation.py`: the hardcoded `MODELS`/`RUNS`/`SITES` and the `{folder}/{model}_{run}/{mode}/{site}/` layout tie it to the 6x6 study tree, and `load_prediction` raises `FileNotFoundError` on the first missing combination of 210. Fine as a one-off — worth a module docstring saying so, so it is not mistaken for a general tool.

Ultimate-Storm (approved):
Verified. Merging.

PR #518, opened 2026-08-31
ODELIA docker build: overwrite master_template.yml only after NVFlare installation 

oleschwen:
When building the ODELIA Docker image, improve docker build cache usage by copying our custom master_template.yml to the installed NVFlare directory so that we can re-use the cached result of installing the NVFlare python package.
(pulled out from https://github.com/KatherLab/MediSwarm/pull/391)

### DIFF
--- a/docker_config/Dockerfile_ODELIA
+++ b/docker_config/Dockerfile_ODELIA
@@ -286,9 +286,9 @@ RUN chmod a+rwX /huggingface_home -R
 # install ODELIA fork of NVFlare from local source
 WORKDIR /workspace/
 COPY ./MediSwarm/docker_config/NVFlare /workspace/nvflare
-## Override with MediSwarm master_template.yml (contains version placeholders injected by build script)
-COPY ./MediSwarm/docker_config/master_template.yml /workspace/nvflare/nvflare/lighter/templates/
 RUN python3 -m pip install --no-cache-dir /workspace/nvflare && rm -rf /workspace/nvflare
+## Override with MediSwarm master_template.yml (contains version placeholders injected by build script)
+COPY ./MediSwarm/docker_config/master_template.yml /opt/conda/lib/python3.10/site-packages/nvflare/lighter/templates/master_template.yml
 
 # ---------------------------------------------------------------------------
 # Stage 5: application code (changes every commit)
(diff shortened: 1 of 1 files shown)


Ultimate-Storm:
Correct, and the cache win is real: `buildDockerImageAndStartupKits.sh:74-75` seds the version into `master_template.yml` on every build, so the old COPY position invalidated the pip-install layer each time. Destination verified against 1.6.0 — it matches `nvflare.lighter.__file__`.

One fragility: the path hardcodes `python3.10`. On a Python bump, COPY creates the directory instead of failing and kits ship with unstamped placeholders. Worth asserting the hardcoded path is where nvflare actually lives:

```dockerfile
RUN python3 -c "import nvflare.lighter,os;assert os.path.dirname(nvflare.lighter.__file__)=='/opt/conda/lib/python3.10/site-packages/nvflare/lighter'"
```

oleschwen:
Good point, we should not hard-code python3.10.
However, shouldn't we copy the file to the intended location and avoid the hard-coded path altogether?

Ultimate-Storm (approved):
Verified. Merging.

Ultimate-Storm:
Not this PR, and not a file-permission issue.

The build fails at `Dockerfile_STAMP:58`:

```
RUN python3 -m pip install --no-cache-dir "stamp @ git+https://github.com/KatherLab/STAMP.git@2.4.0"
```

pip shells out to `git clone`, GitHub refuses the unauthenticated request, git falls back to prompting for credentials and dies with no tty:

```
fatal: could not read Username for 'https://github.com': No such device or address
fatal: expected flush after ref listing
```

`expected flush after ref listing` is the tell — GitHub returned a non-protocol response during ref advertisement, i.e. throttling, not authorization.

Evidence it is environmental rather than yours:

- The identical failure hit `fix/503-total-rounds-single-source` at 09:15 the same morning (run 33613078715). That branch touches neither Dockerfile.
- `fix/395-supply-chain` passed at 08:39 that morning; this branch passed on 08-31.
- `KatherLab/STAMP` is public, returns HTTP 200 unauthenticated, and `git ls-remote` for tag 2.4.0 succeeds right now from cosmos **and** from both dl0 and dl2.

So: re-running should clear it.

For the recurring case, the durable fix is to stop using a git clone for a pinned tag and fetch the release tarball instead — plain HTTPS, no git, no credential path:

```dockerfile
RUN python3 -m pip install --no-cache-dir \
    "stamp @ https://github.com/KatherLab/STAMP/archive/refs/tags/2.4.0.tar.gz"
```

Verified pip resolves and builds metadata from that URL (my throwaway test only failed on Python 3.10 vs STAMP's `>=3.11`; this image is 3.11, so it applies cleanly). Happy to open that as a separate PR — it does not belong in this one.

Issue #526, opened 2026-09-03
D2.5 (T2.4): compare regionally fine-tuned models against the pan-European baseline


Ultimate-Storm:
Grant task T2.4 "Implement regional fine-tuning of AI models" (M28–M48). Deliverable **D2.5**, due **M48 (Dec 2026)**, type R — report, PU.

The grant names the experiment explicitly: compare the best pan-European model against models fine-tuned to the **Dutch** and **Greek** populations.

## Plan

- Fine-tune the `87c5bbee` global checkpoint (1DivideAndConquer) on:
  - **Dutch cohort** — RUMC_1 + UMCU_1
  - **Greek cohort** — MHA_1
- Evaluate each against the pan-European baseline on the same held-out split, using the same evaluation code.

## Baselines already measured

From the external validation of `87c5bbee` on the held-out ODELIA challenge test split (macro AUROC / lesion-vs-none / malignant-vs-rest):

| Site | n | Macro | Lesion | Malignant |
|---|---|---|---|---|
| RUMC_1 | 8 | 0.857 | 0.857 | 0.857 |
| UMCU_1 | 41 | 0.602 | 0.684 | 0.750 |
| MHA_1 | 26 | 0.815 | 0.931 | 0.955 |
| pooled | 165 | 0.820 | 0.853 | 0.887 |

No baseline re-run needed. Note UMCU is the weakest centre — the Dutch cohort is where regional fine-tuning has the most room to show an effect, and RUMC at n=8 is indicative only.

## Reuse, do not rewrite

- `scripts/evaluation/predict.py` — `--checkpoint --model-name 1DivideAndConquer --num-classes 3 --split test`
- `workspace/eval_87c5bbee/analyze.py` — pooling and AUROC via Mann-Whitney U

## Done when

A comparison report exists showing fine-tuned vs pan-European per region on the same cohort and the same evaluation code. A null result is a valid D2.5 — what is not valid is comparing across different cohorts or different metric code.

Depends on #525 for per-region metric collection.

Ultimate-Storm:
## Experiment design, and a power constraint worth settling first

Started this on `feat/per-site-metrics` (#534), since D2.5 needs the per-region metrics that PR restores.

### The constraint

Regional test partitions from the 8-site run (`87c5bbee`, best checkpoint), counted from the per-case predictions rather than by summing per-site AUROCs:

| Cohort | n | No lesion | Benign | Malignant |
|---|---:|---:|---:|---:|
[...]

PR #533, opened 2026-09-04
docs: record three evaluation pitfalls that produced wrong conclusions

Ultimate-Storm:
Companion to `SWARM_FAILURE_MODES.md` — that covers how a *run* fails, this covers how a *number* misleads. All three entries actually happened on this project.

**E1 — macro AUROC does not rank centres at small class support.** The per-site table ranked UMCU worst (0.602) and RUMC best (0.857). Both were artefacts: the ordering tracks how many benign cases each test split happened to contain (0, 1, 2, 2, 12), because macro AUROC gives that class a third of the weight regardless of support.

| Site | n | benign | benign AUROC | macro | macro w/o benign |
|---|---|---|---|---|---|
| UMCU_1 | 41 | 2 | 0.372 | 0.602 | **0.717** |
| CAM_1 | 64 | 2 | 0.315 | 0.710 | **0.908** |
| RUMC_1 | 8 | 0 | n/a | 0.857 | 0.857 |

CAM was penalised as hard as UMCU and nobody noticed. RUMC "led" only because two classes were averaged instead of three. It is a split artefact — benign is ~10% of UMCU's challenge slice overall, but only 2 of 20 landed in the 41-case test partition.

Also records that these are the **challenge test partition**, not any site's training data — 41 of 204 curated exams against a >10,000-exam training contribution. Without that, "UMCU's results are weak" reads as a statement about their data.

**E2** — checkpoint loaded with the wrong architecture loads silently and yields meaningless-but-plausible metrics.

**E3** — accuracy hid a mis-set decision threshold that under-called malignancy.

Flags #526 (D2.5 regional fine-tuning) as the next place E1 applies, since it compares per-region performance on these same small partitions.

### DIFF
--- a/docs/EVALUATION_PITFALLS.md
+++ b/docs/EVALUATION_PITFALLS.md
@@ -1,0 +1,81 @@
+# ODELIA Evaluation — Pitfalls
+
+Companion to `SWARM_FAILURE_MODES.md`. That document covers ways a *run* fails; this one covers
+ways a *number* misleads. Every entry here actually happened on this project and cost real time or
+produced a wrong conclusion.
+
+## Quick reference
+
+| | Pitfall | Tell |
+|---|---|---|
+| **E1** | Macro AUROC used to rank centres at small class support | One class has single-digit positives |
+| **E2** | Checkpoint evaluated with the wrong architecture | Loads silently; metrics look plausible but are meaningless |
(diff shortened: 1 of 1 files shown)


oleschwen (approved):
Useful summary, no change to either swarm learning or application code.

PR #534, opened 2026-09-04
feat(metrics): publish per-site metrics with class support (#525, #441)

### LINKED ISSUE #441: ODELIA: Align metric computation during and after training
Following #50 and #410, we should align the computation of metrics computed and logged during training with the metrics we can compute from logged predictions, possibly extending the logging to include raw logits, and avoiding duplicate implementation (if any).

### LINKED ISSUE #525: Swarm p2p runs report no per-site metrics to the coordinator
Blocks T2.4, T3.2, T3.3 and T3.4. Doing this first is the cheapest way to unblock all four.

## What happens

`cross_site_val/cross_val_results.json` in the server-side job workspace is `{}` after every peer-to-peer swarm run. Nothing in `application/` or `scripts/` writes to it.

Confirmed on two runs at opposite ends of the scale:

| Job | Run | Result |
|---|---|---|
| `87c5bbee` | 8 sites, 20 rounds, 1 d 19 h 53 m | `{}` |
| `85159a6d` | 8 sites, 1 round, 52 seconds | `{}` |

Per-site validation figures currently have to be gathered by logging into individual hospitals by hand. A published result cannot depend on that.

## Why it is the critical path

- **T3.2 regulatory testing node** exists specifically to receive per-site performance and disease-incidence figures from training nodes. There is nothing to receive them from.
- **D2.5** (T2.4) needs per-region metrics to compare regionally fine-tuned models against the pan-European baseline.
- **D3.5** (T3.4) needs a performance comparison against the WP1 and WP2 results.
- **T3.3 Phase 2** needs per-round epsilon logged alongside utility metrics.

## Reproduction — one minute

```
submit_job /fl_admin/local/mediswarm_jobs/minimal_training_pytorch_cnn_strict8_retry_20260811_1527
# then, in the server container:
cat /tmp/nvflare/jobs-storage/<job_id>/workspace  -> unzip -> cross_site_val/cross_val_results.json
```

Expect `{}`. That job is strict 8-of-8, 1 round, and finishes in 52 seconds, so this is a fast iteration loop.

## Done when

A 1-round strict-8 smoke run leaves a non-empty `cross_val_results.json` with one entry per participating site.

Related: #441 (which metrics are computed — adjacent but distinct from whether they are transported).

Ultimate-Storm:
Closes #525. Closes #441.

Treated as one piece of work: per-site results never reached the coordinator, and the results we *did* gather by hand were reported in a shape that misled us. Fixing transport without fixing shape would have industrialised the second problem.

## Transport (#525)

Clients already stream metrics server-ward — `MetricRelay` fires `fed.analytix_log_stats`, `ClientFedEventRunner` relays it, `ServerFedEventRunner` delivers it. **Nothing on the server ever handled that event**, so `cross_site_val/cross_val_results.json` came back `{}` after every peer-to-peer run — the two-day 20-round run `87c5bbee` and a 52-second smoke run alike.

`PerSiteMetricCollector` closes the last hop: it accumulates the stream per site and, at end of run, hands it to the `ValidationJsonGenerator` **already configured in every server config**. No client change, no new transport, no image or kit rebuild — it ships with the job via `_shared/custom`.

## Shape (#441)

`compute_epoch_metrics` now emits `{state}/support_class{i}` and `{state}/n` at every tier.

A per-class metric read without its support is what produced a wrong conclusion about UMCU. Macro AUROC ranked centres by how many benign cases their split happened to contain, not by performance:

| Site | n | benign cases | benign AUROC | macro | macro w/o benign |
|---|---|---|---|---|---|
| UMCU_1 | 41 | 2 | 0.372 | 0.602 | **0.717** |
| CAM_1 | 64 | 2 | 0.315 | 0.710 | **0.908** |
| RUMC_1 | 8 | 0 | n/a | 0.857 | 0.857 |

CAM was penalised just as hard and nobody noticed; RUMC "led" only because two classes were averaged instead of three. Recorded as E1 in `docs/EVALUATION_PITFALLS.md` (#533).

## Choices worth reviewing

- Support values are **0-dim tensors**, keeping the returned mapping uniform — an existing test asserts `v.ndim == 0` for every value.
- They are emitted **before** the extended-tier early return, so support travels whether or not that tier is on.
- Two existing assertions were **reconciled, not weakened**: the tier gate is about per-class *metrics*, and train state now carries support too.
- `minimal_training_pytorch_cnn` gets a single-file symlink. It is deliberately self-contained, but the collector is job-agnostic and this is what makes a **52-second** verification of the transport possible instead of a multi-hour one.

## Verification

**Live, on all eight sites.** Job `c5520641`, 34 seconds, `FINISHED:COMPLETED`:

```json
{"MHA_1": {"swarm_global": {"val/loss_epoch": 0.6587, "val/ACC": 0.5, "val/AUC_ROC": 1.0}},
 "VHIO_1": {...}, "USZ_1": {...}, "RUMC_1": {...},
 "RSH_1": {...}, "UKA_1": {...}, "UMCU_1": {...}, "CAM_1": {...}}
```

All eight report identical values because `MiniDatasetForTesting` is deterministic — ten entries derived from the index, split with `random_state=42` — so every site holds byte-identical data. That confirms the transport, not the science.

### The live run found a second bug

The first live attempt still returned `{}`, and the server log said the metrics had arrived:

```
PerSiteMetricCollector - INFO - Published metrics for 8 site(s):
    CAM_1, MHA_1, RSH_1, RUMC_1, UKA_1, UMCU_1, USZ_1, VHIO_1
```

Every hop worked. The failure was ordering *within a single event*: `ValidationJsonGenerator` dumps the JSON in its own `END_RUN` handler, `AnalyticsReceiver.finalize` also runs on `END_RUN`, and components are invoked in the order they appear in `config_fed_server.conf`. `json_generator` is listed first in all seven job configs, so the file was written and only then did the collector publish into its in-memory dict.

It read as a missing metric rather than a race because `ServerRunner` logs `END_RUN fired` *after* the dispatch returns — the publish line appears ~2 s before it.

Fixed by publishing on `ABOUT_TO_END_RUN`, which is fired strictly earlier, so the result no longer depends on how any job config happens to be ordered. `END_RUN` stays as a fallback and `finalize` is idempotent. Recorded as **F10** in `docs/SWARM_FAILURE_MODES.md`.

### CI was green the whole time

`pytest.importorskip("nvflare")` **skipped `test_per_site_metrics.py` entirely** — the unit-tests workflow never installed NVFlare. The same trap the `torchio`/`pandas` lines were added for (#416/#423). The workflow now installs the in-tree fork, so CI tests the same NVFlare the image ships. A skipped test file is not a passing one.

16 tests, all running. Full suite locally: **378 passed, 7 skipped**.

### DIFF
--- a/application/jobs/_shared/custom/threedcnn_ptl.py
+++ b/application/jobs/_shared/custom/threedcnn_ptl.py
@@ -582,6 +582,26 @@ def prepare_training(logger, max_epochs: int, site_name: str = None,
         # compared to pure SGD updates every sample.
         accumulate_grad_batches = 8
 
+        # Metrics reach the coordinator only if something writes them to
+        # NVFlare's pipe. TensorBoardLogger writes to local disk, which is why
+        # per-site results had to be collected by logging into each hospital by
+        # hand and cross_val_results.json came back empty (#525). ClientLogger
+        # forwards the same self.log() calls through the pipe, where MetricRelay
+        # turns them into fed.analytix_log_stats and PerSiteMetricCollector
+        # publishes them server-side.
+        #
+        # Only meaningful inside a swarm run: outside one there is no client
+        # runtime to write to, so the local TensorBoard logger stands alone.
+        trainer_loggers = [TensorBoardLogger(save_dir=path_run_dir)]
+        if os.environ.get("TRAINING_MODE", "") == "swarm":
+            try:
+                from nvflare.app_opt.lightning.loggers.client_logger import ClientLogger
+                trainer_loggers.append(ClientLogger())
+                logger.info("ClientLogger attached: metrics will stream to the coordinator")
+            except Exception as exc:
+                # A missing metric stream must not stop a training run.
+                logger.warning(f"ClientLogger unavailable, metrics stay local: {exc}")
+
         trainer = Trainer(
             accelerator='gpu',
             accumulate_grad_batches=accumulate_grad_batches,
@@ -594,7 +614,7 @@ def prepare_training(logger, max_epochs: int, site_name: str = None,
             log_every_n_steps=log_every_n_steps,
             max_epochs=max_epochs,
             num_sanity_val_steps=2,
-            logger=TensorBoardLogger(save_dir=path_run_dir)
+            logger=trainer_loggers
         )
 
     except Exception as e:
--- a/.github/workflows/unit-tests.yaml
+++ b/.github/workflows/unit-tests.yaml
@@ -51,6 +51,13 @@ jobs:
           # fold validation). Without them pytest silently SKIPS those files, which is
           # how a broken guard test reached main (#416/#423).
           pip install torchio pandas
+          # The NVFlare fork gates the server-side metric collector tests
+          # (test_per_site_metrics.py). Same trap as torchio/pandas above: without
+          # it pytest SKIPS the file and the job still reports green, which is how
+          # an empty cross_val_results.json survived two live 8-site runs (#525).
+          # Installed from the in-tree fork so CI tests the same NVFlare the image
+          # ships, not whatever PyPI currently calls 'nvflare'.
+          pip install ./docker_config/NVFlare
 
       - name: Run unit tests
         run: |
--- a/application/jobs/ODELIA_ternary_classification/app/config/config_fed_server.conf
+++ b/application/jobs/ODELIA_ternary_classification/app/config/config_fed_server.conf
@@ -8,6 +8,14 @@ components = [
     path = "nvflare.app_common.widgets.validation_json_generator.ValidationJsonGenerator"
     args {}
   }
+  {
+    # Collect the per-site metrics clients already stream (fed.analytix_log_stats)
+    # and hand them to json_generator at end of run, so cross_val_results.json is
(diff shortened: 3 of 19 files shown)


application/jobs/_shared/custom/threedcnn_ptl.py
```diff
@@ -582,6 +582,26 @@ def prepare_training(logger, max_epochs: int, site_name: str = None,
...
+        # forwards the same self.log() calls through the pipe, where MetricRelay
+        # turns them into fed.analytix_log_stats and PerSiteMetricCollector
+        # publishes them server-side.
+        #
+        # Only meaningful inside a swarm run: outside one there is no client
+        # runtime to write to, so the local TensorBoard logger stands alone.
+        trainer_loggers = [TensorBoardLogger(save_dir=path_run_dir)]
+        if os.environ.get("TRAINING_MODE", "") == "swarm":
+            try:
+                from nvflare.app_opt.lightning.loggers.client_logger import ClientLogger
```
oleschwen: Is this try-catch necessary (same in the minimal training)?
We don't need to dynamically distinuish what is available at runtime, we know whether or not this is available when building the image and in a test run, and move the import to the beginning of the file.

Issue #541, opened 2026-09-08
Swarm runs do not verify that all sites resolved the same MODEL_NAME


Ultimate-Storm:
## Problem

For swarm training the model is chosen **per site, at client start**, not by the submitted job, and nothing checks that the sites agree.

`ODELIA_ternary_classification` is generic by design: its launcher line sets no `MODEL_NAME`, and its persistor is `models.models_config.create_model`. Both the trainer and the persistor fall back to the `MODEL_NAME` environment variable, which `docker.sh` always supplies:

```
--env MODEL_NAME=${CLI_MODEL_NAME:-${MODEL_NAME:-MST}}      # docker.sh:164
```

That is a useful property — it means ResNet and Swin3D work for swarm runs without editing `config_fed_client.conf`. It also means a site that starts its client with `--model_name Swin3D` while the others default to `MST` will join a run and contribute weights from a different architecture. The aggregation then combines mismatched parameter tensors.

The challenge jobs are not exposed: each pins `MODEL_NAME=<team>` in its own launcher line and a team-specific persistor.

## Why this needs a guard rather than documentation

There is no point in the run where the mismatch is announced. Each client logs its own model name locally and proceeds. Whether the aggregation errors or silently produces a meaningless global depends on how the shapes happen to line up — and a silent success is the bad case, because the resulting model would be reported as a normal result.

This is the same class of hazard as #535 (warm-start checkpoint with no provenance) and E2 in `docs/EVALUATION_PITFALLS.md` — "wrong architecture loads silently".

STAMP already has the equivalent guard for a different property: #519 verifies cross-site class order before a run starts.

## Suggested fix

- [ ] Have each client report its resolved `MODEL_NAME` during configuration, and have the controller refuse to start when they are not identical — mirroring the STAMP class-order check.
- [ ] Record the resolved model name in the run's metadata so a completed run states which architecture produced it.
- [ ] Until that exists, note in the partner docs that `--model_name` must match across all sites for a swarm run.

## Status

Untested. Raised from a code reading prompted by a question about how to run non-default models; I have not deliberately started a mismatched run to see what happens. Worth establishing that before relying on ResNet/Swin3D for a consortium run.

PR #545, opened 2026-09-08
Label warm-start checkpoints and refuse another model's weights (#535)

### LINKED ISSUE #535: Warm-start mirror is shared across all architectures and has no provenance guard
All six real ODELIA jobs read and write the same warm-start mirror:

```
source_ckpt_file_full_name = "/scratch/mediswarm_latest_global.pt"
latest_global_path         = "/scratch/mediswarm_latest_global.pt"
```

`challenge_1DivideAndConquer`, `challenge_2BCN_AIM`, `challenge_3agaldran`, `challenge_4abmil`, `challenge_5pimed` and `ODELIA_ternary_classification` — **five different architectures, one file**. `warm_continue.py` records no architecture, model name or provenance alongside the checkpoint, and checks none on load.

Two consequences.

**1. Any run silently replaces the previous global at every site.** Verified today: before job `24cdf247` the mirror on UKA was 722,229,662 bytes dated 15 Aug — `sha256 ad26080c71161e6f8abb6a279c06a41d459b7bb237699cbe2d4522ddb942270d`, byte-identical to the `87c5bbee` pan-European global. A one-round run overwrites it. Nothing warns, and the previous checkpoint is only recoverable if someone happened to collect it centrally beforehand.

**2. A cross-architecture warm start is not guarded.** Run 2BCN_AIM, then 1DC, and the 1DC run warm-starts from a 2BCN_AIM checkpoint. Whether that raises or silently loads a partial state dict, neither outcome is acceptable: this is E2 in `docs/EVALUATION_PITFALLS.md` — "wrong architecture loads silently" — with the pathway wired in by default.

This matters concretely for D2.5 (#526), where regional fine-tuning must start from *exactly* the checkpoint the pan-European baseline metrics were computed from.

### Suggested fix

- Namespace the mirror by model: `/scratch/mediswarm_latest_global_{MODEL_NAME}.pt`.
- Write a sidecar (model name, job id, round, sha256) and refuse to warm-start when the model name does not match, rather than trusting the tensor shapes.
- Keep the previous mirror as `.prev` so one unintended run is not destructive.

Found while preparing #526. Not urgent for correctness of any completed run — `87c5bbee` is preserved centrally and checksum-verified — but it is a live trap for every future one.

Ultimate-Storm:
Closes #535.

All six ODELIA jobs share one warm-start mirror path and five architectures can land there. Nothing recorded which model wrote a checkpoint, and nothing checked on load — so running 2BCN_AIM then 1DivideAndConquer warm-starts the second from the first's weights. Shape compatibility is not a safe proxy for correctness; this is **E2** in `docs/EVALUATION_PITFALLS.md`, with the pathway wired in by default.

This also came within one round of destroying the pan-European baseline **three separate times** in the week of 4–8 Sep, each time caught by hand.

## What changes

Every mirrored checkpoint gets a `.provenance.json` sidecar (model name, job id, sha256), and `load_model` refuses one whose recorded model differs from `MODEL_NAME`.

## Two choices worth reviewing

**Unlabelled checkpoints are accepted with a warning, not refused.** Every mirror written before this change has no sidecar; failing those would break warm-continue at every site for no safety gain.

**Provenance is written separately from the copy.** My first version put it inside the copy's `try`, where a provenance failure made a *successful* mirror log as failed — caught by the existing `test_warm_continue` tests, and now pinned by its own regression test. The checkpoint matters; its label is a safety net and must not be able to damage it.

The digest streams in 1 MB blocks — these files are ~700 MB.

## Tests

10 new, full suite **369 passed, 7 skipped**.

Related: #541 (no cross-site check that all sites resolved the same MODEL_NAME) is the same hazard at a different layer.

### DIFF
--- a/application/jobs/_shared/custom/warm_continue.py
+++ b/application/jobs/_shared/custom/warm_continue.py
@@ -29,6 +29,8 @@ auto-resume: each block continues from the previous block's global. Collect the
 mirrors from all sites with ``scripts/collect_swarm_globals.sh`` (#160).
 """
 
+import hashlib
+import json
 import os
 import shutil
 
@@ -49,6 +51,55 @@ WARM_START_MODES = {
     WARM_START_MODE_REQUIRE,
 }
 WARM_START_REQUIRED_MISSING = "WARM_START_REQUIRED_MISSING"
+WARM_START_MODEL_MISMATCH = "WARM_START_MODEL_MISMATCH"
+
+
+
+def sidecar_path(ckpt_path: str) -> str:
+    """Provenance file that travels with a mirrored checkpoint."""
+    return f"{ckpt_path}.provenance.json"
+
+
+def current_model_name() -> str:
+    """The model this process is training, as the job's launcher set it."""
+    return os.environ.get("MODEL_NAME", "") or "unknown"
+
+
+def read_provenance(ckpt_path: str):
+    """Return the recorded provenance dict, or None if there is none."""
+    try:
+        with open(sidecar_path(ckpt_path)) as fh:
+            return json.load(fh)
+    except Exception:
+        return None
+
+
+def write_provenance(ckpt_path: str, model_name: str, job_id: str = "", digest: str = ""):
+    try:
+        with open(sidecar_path(ckpt_path), "w") as fh:
+            json.dump(
+                {"model_name": model_name, "job_id": job_id, "sha256": digest},
+                fh,
+            )
+    except Exception:
+        # provenance is a guard, not a feature -- never fail a run over it
+        pass
+
+
+def file_digest(path: str) -> str:
+    """sha256 of a checkpoint, or "" if it cannot be read.
+
+    Read in blocks: these files are ~700 MB and must not be loaded into memory
+    just to be identified.
+    """
+    try:
+        h = hashlib.sha256()
+        with open(path, "rb") as fh:
+            for block in iter(lambda: fh.read(1024 * 1024), b""):
+                h.update(block)
+        return h.hexdigest()
+    except Exception:
+        return ""
 
 
 def normalize_warm_start_mode(warm_start_mode: str) -> str:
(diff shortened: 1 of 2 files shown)


application/jobs/_shared/custom/warm_continue.py
```diff
@@ -49,6 +51,55 @@
...
+
+def write_provenance(ckpt_path: str, model_name: str, job_id: str = "", digest: str = ""):
+    try:
+        with open(sidecar_path(ckpt_path), "w") as fh:
+            json.dump(
+                {"model_name": model_name, "job_id": job_id, "sha256": digest},
+                fh,
+            )
+    except Exception:
+        # provenance is a guard, not a feature -- never fail a run over it
```
oleschwen: I'd log a warning (error?) in this case instead of silently continuing.

application/jobs/_shared/custom/warm_continue.py
```diff
@@ -49,6 +51,55 @@
...
+    Read in blocks: these files are ~700 MB and must not be loaded into memory
+    just to be identified.
+    """
+    try:
+        h = hashlib.sha256()
+        with open(path, "rb") as fh:
+            for block in iter(lambda: fh.read(1024 * 1024), b""):
+                h.update(block)
+        return h.hexdigest()
+    except Exception:
```
oleschwen: I'd log this as an error or even omit the try/except – or are there valid reasons why requesting a hash of a file should be allowed to fail?

oleschwen:
Accepting checkpoints without sidecar sounds fine as long as we want to continue with older checkpoints. I'm not sure how long that is the case, we could consider dropping this backwards compatibility at some point.

Ultimate-Storm:
Agreed, and I'd rather it were stricter too — my first instinct was to refuse unlabelled checkpoints outright.

The reason I didn't: every mirror currently sitting at `/scratch/mediswarm_latest_global.pt` on all eight sites was written before this change, so none has a sidecar. Refusing them would break warm-continue everywhere at once, and at several sites we have no shell access to place a checkpoint by hand — so there'd be no way back except a full run from scratch.

A clean way to drop the compatibility: once every site has completed one run on a version carrying this change, every live mirror has a sidecar, and unlabelled can become a hard failure. That's checkable rather than a guess — the sidecar is written on every mirror, so `ls /scratch/mediswarm_latest_global.pt.provenance.json` at each site answers it.

Happy to add a `warm_start_strict_provenance` flag now, defaulting to permissive, so the switch is a config change rather than a code change when we're ready. Say the word and I'll add it to this PR.

oleschwen:
I'd suggest waiting with the implementation until we can actually use it.

Issue #528, opened 2026-09-03
D3.3 (T3.2): SL software v2.0 — active learning plus regulatory testing node


Ultimate-Storm:
Grant task T3.2 (M31–M48 for the work; deliverable **D3.3** due **M54 (Jun 2027)**).

The software half of T3.2, separated from the D3.2 report (#527) because they have different deadlines.

## Regulatory testing node

Grant text: "a dedicated testing node will also be included that allows for reception of more detailed information from the respective training nodes such as performance of the network on local training and testing sets or incidence rates of disease occurrences, to close the gap between incoming information and fast and actionable insights."

The worked example in the grant is outbreak monitoring — a spike in the rate of AI-positive cases at regional sites as an early signal.

**This depends entirely on #525.** The node exists to receive per-site performance and incidence figures, and today nothing transports them: `cross_val_results.json` is `{}` after every p2p run. Build #525 first or there is nothing for the node to consume.

## Also in scope

- Promote the D3.2 active-learning experiment (#527) into a supported feature rather than an experiment script.

## Done when

A testing-node participant can join a run and receive per-site performance and incidence figures without anyone logging into a hospital by hand.

Ultimate-Storm:
**Status 14 Sep 2026 (M45) — 1 of 3**
- [x] Per-site metrics transport (#525): the testing node has something to consume
- [ ] Testing-node participant that receives per-site performance and incidence figures
- [ ] Active learning promoted from experiment (#555, #563) to a supported feature

Issue #527, opened 2026-09-03
D3.2 (T3.2): demonstrate the usefulness of active learning


Ultimate-Storm:
Grant task T3.2 (M31–M48). Deliverable **D3.2**, due **M48 (Dec 2026)**, type R — report, PU.

Grant text: "the image types and label categories in which the model has a low performance will be actively requested by the model and can then be supplied for the next training iteration."

Nothing on active learning exists in the repo today — this is greenfield.

## Scope for M48

This deliverable is a **report demonstrating usefulness**, not production software. The software half is D3.3 at M54 (#528). Keep this first version simple.

- Uncertainty-based selection that requests the low-performing label categories.
- Measure against a **random-selection control on the same labelling budget**. "The model asked for data" is not on its own a demonstration of usefulness — the deliverable needs a budget curve.
- Prime candidate signal: the model currently under-calls malignancy (predicts malignant 21× where truth is 37, on the challenge set) while ranking it well (AUROC 0.887). That is a concrete low-performance category to request against.

## Round-cost budget — this governs feasibility

Measured from `87c5bbee`: 20 rounds in 1 d 19 h 53 m ≈ **2.2 h per round** for 1DC at 8 sites, gated by UKA.

So 5 AL iterations × 5 rounds ≈ **55 h per configuration**, and the report needs at least two configurations (AL and the random control). Design to fit, or it will not land by M48. The 1-round strict-8 smoke job (52 s) is available for plumbing checks before committing to a real run.

## Fleet status

The 8-site fleet is live and verified (8/8 registered and configured as of 2026-09-03), and partners have already been asked to keep clients online for exactly this work.

## Done when

A report showing labelling-budget curves for active vs random selection on the same cohort.

Ultimate-Storm:
## Design, costed against the measured round budget

Nothing on active learning exists in-tree, so this is greenfield. The binding constraint is compute, not ideas.

**Measured, not estimated:** the completed 8-site run did 20 rounds in 1 d 19 h 53 m — **2.2 h per round**, gated by the slowest site. `compute_weighted_epochs` gives every ODELIA site **1 epoch per round** (`5 × 500/6133 = 0.4`, clamped to 1), so a round is one pass over local data plus two model transfers.

[...]

PR #555, opened 2026-09-08
feat(active-learning): sample-selection strategies for D3.2 (#527)

Ultimate-Storm:
Refs #527. First code for D3.2; nothing on active learning existed in-tree.

## Why this is measured on prediction rows, not swarm rounds

D3.2 asks whether uncertainty-based selection reaches a given performance with **fewer labels** than random. That is a property of the selection rule and the data.

The federated form costs about **110 h of consortium time** across two arms — 2.2 h/round measured on the completed 8-site run, and `compute_weighted_epochs` gives every ODELIA site 1 epoch per round. That does not fit before M48 alongside D2.5 and D3.4, on a fleet that currently cannot hold a long run together. One federated confirmation run on the winning strategy demonstrates it in the real system for roughly a tenth of the cost.

## What's here

`entropy`, `margin` (top-1 minus top-2) and `random` as the control. Margin is returned as `1-(top1-top2)` so both measures point the same way — otherwise every caller has to remember that a *small* margin means *high* uncertainty.

## Two choices driven by how the data actually arrives

**Ties break on uid, not input order.** These CSVs come in whatever order each site's dataloader produced; an order-dependent tie-break would make the same experiment give different answers at different sites. Pinned by a test that reverses the input.

**Degenerate probability rows normalise rather than raise.** Six model implementations write these files. A row that doesn't sum to 1 is a data problem, not a reason to abort a study.

## The function that exists to stop a wrong conclusion

`class_composition()` is included because the analysis must not skip it. Uncertainty sampling over-selects rare classes, and ODELIA's benign class is rare at every site — so a gain over random has to survive controlling for class mix, or the experiment has measured class imbalance and called it active learning. The test shows entropy taking **all three** benign cases out of a 20:3 pool.

This is the same trap as E1 in `docs/EVALUATION_PITFALLS.md`, which has already produced one wrong conclusion about a partner site.

## Tests

12 new; full suite **419 passed, 7 skipped**.

## Dependency

The study needs per-case probabilities returned from each site — the same plumbing #526 (D2.5) is blocked on and #528 (D3.3 testing node) requires.

### DIFF
--- a/scripts/active_learning/selection.py
+++ b/scripts/active_learning/selection.py
@@ -1,0 +1,117 @@
+#!/usr/bin/env python3
+"""Sample-selection strategies for the active-learning study (D3.2 / T3.2, #527).
+
+What this is for
+----------------
+D3.2 asks whether active learning is *useful*: whether selecting which cases to
+label by model uncertainty reaches a given performance with fewer labels than
+selecting at random. That is a property of the selection rule and the data, so
+it is measured here on per-case class probabilities rather than by running a
+swarm round per iteration -- 2.2 h per round measured on the 8-site run makes
+the federated form of this experiment cost ~110 h across two arms, which does
+not fit before M48.
+
+Strategies
+----------
+``entropy``  predictive entropy over the class distribution. Highest when the
+             model spreads mass evenly; picks cases it cannot commit on.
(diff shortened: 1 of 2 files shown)


oleschwen (approved):
Looks good, does not affect the swarm learning or application code.

Issue #529, opened 2026-09-03
D3.4: measures needed to make SL-based AI robust against adversarial attacks


Ultimate-Storm:
Deliverable **D3.4**, due **M48 (Dec 2026)**, WP3, type R — report, PU.

Description from the grant: "Report describing potential adversarial attacks and the subsequent measures needed to make SL-based AI fully robust against such attacks."

## Keep this distinct from differential privacy

D3.4 is about **adversarial robustness**. T3.3 (#530) and MS6 (#531) are about **privacy** — noise, epsilon budgets, and reconstruction of training data. They are separate deliverables under the same work package and it is easy to let one absorb the other. Do not.

The white-hat attack in MS6 tests whether personal information can be reconstructed from model weights. D3.4 covers attacks against the model's behaviour.

## Done when

The report exists and covers the attack surface of a swarm setting specifically — not a generic adversarial-ML literature review. Sites exchange weights, and any site can be an adversary; that is what makes this different from the single-institution case.

Ultimate-Storm:
## Starting point, from the code rather than from the grant text

Every one of the nine job configurations aggregates with:

```
nvflare.app_common.aggregators.intime_accumulate_model_aggregator.InTimeAccumulateWeightedAggregator
```

That is a **plain weighted mean**, and the fork ships no alternative — `nvflare/app_common/aggregators/` contains accumulate, collect-and-assemble, DXO and lazy variants, and nothing robust. No trimmed mean, no coordinate-wise median, no Krum.

The consequence is worth stating plainly for the deliverable: **a single participant can move the global model arbitrarily.** A mean has an asymptotic breakdown point of zero — one update scaled large enough dominates the other seven, no matter how well the rest behave. That is not a hypothesis about our deployment; it follows from the aggregation rule in use.

This is the distinctive swarm-learning threat and belongs at the centre of D3.4. Input-space attacks (FGSM/PGD on images) apply to any deployed model and are the better-covered literature; a compromised or malfunctioning *participant* is what swarm learning adds.

## Proposed structure

**1. Threat model, written down.** Who can do what: an honest-but-curious site, a site with a corrupted pipeline, a compromised client host. Distinguish deliberate attack from the failure we have actually seen — a site returning garbage because its host crashed mid-round.

[...]

PR #556, opened 2026-09-08
feat(metrics): return per-case predictions, opt-in per site (#526, #527, #528)

Ultimate-Storm:
Four deliverables converge on this one capability:

| | needs |
|---|---|
| **D2.5** (#526) | paired per-case scores to compare regional vs pan-European |
| **D3.2** (#527) | per-case scores to rank cases by uncertainty |
| **D3.3** (#528) | *is* a node that receives per-site figures |
| **D3.4** (#529) | case-level results for the robustness analysis |

The alternative was moving a 722 MB checkpoint out of each hospital — which is what D2.5 is currently blocked on, and which nobody here has the access to do. Summary metrics already travel (`per_site_metrics.py`); this extends the same path to the rows behind them.

## What actually leaves a site

```
row_index,ground_truth,prob_class_0,prob_class_1,prob_class_2
```

**No case identifier, no patient identifier, no image.** The index is a position within that site's own validation split and carries no meaning outside it. Pairing two models — what D2.5 needs — works from the index because the validation loader isn't shuffled, so the n-th row is the same case in both runs.

A real uid would have been easier, and is what the first draft of the consortium slide promised. An index is strictly less information and does the same job, so that's what this sends. **The slide needs updating to match.**

## Off by default

This is patient-derived data even pseudonymised, so returning it is the consortium's decision, not a maintainer's. Nothing is emitted unless a site sets `ODELIA_RETURN_PER_CASE=1` in its own environment. The gate lives inside the emit function so there is exactly one place where that decision is made — tested against `0`, `true`, `yes` and empty, all of which must *not* opt in.

## Three things that would otherwise fail silently

- **Chunks are ordered by sequence on write, not by arrival.** The transport gives no ordering guarantee, and a reordered file breaks the index pairing this exists for.
- **A per-site chunk cap** stops one site exhausting coordinator memory — and a file truncated by that cap logs `INCOMPLETE` rather than looking like a valid result.
- **Publication on `ABOUT_TO_END_RUN`**, same component-ordering reason as #534.

## Tests

11 new; full suite **418 passed, 7 skipped**.

## Not included

The client-side call site in `threedcnn_ptl.py`. This PR is the mechanism and its gate; wiring it into training should land once the consortium has agreed to the return in principle.

### DIFF
--- a/application/jobs/_shared/custom/per_case_predictions.py
+++ b/application/jobs/_shared/custom/per_case_predictions.py
@@ -1,0 +1,201 @@
+"""Return per-case predictions to the coordinator (#526, #527, #528).
+
+Why this exists
+---------------
+Four deliverables converge on the same missing capability. D2.5 needs paired
+per-case scores to compare a regionally fine-tuned model against the
+pan-European one; D3.2 needs them to rank cases by uncertainty; D3.3's
+regulatory testing node is defined by receiving per-site figures; and D3.4's
+analysis needs case-level results to say anything about robustness.
+
+The alternative was moving a 722 MB checkpoint out of each hospital, which is
+what D2.5 is currently blocked on and which nobody here has the access to do.
+Per-site *summary* metrics already travel (see ``per_site_metrics.py``); this
+extends the same path to the rows behind them.
+
+What actually leaves a site
+---------------------------
+One line per validation case:
+
+    row_index,ground_truth,prob_class_0,prob_class_1,prob_class_2
+
+**No case identifier, no patient identifier, no image.** The row index is a
+position within that site's own validation split and carries no meaning
+outside it. Pairing two models' predictions -- which is what D2.5 needs --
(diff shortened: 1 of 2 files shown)


oleschwen:
I see some pitfalls here:
* Pairing data across tables assumes that the data is processed in the same order; this may be the case in different rounds of the same training and if the data loader does not change the order, but is this guaranteed?
* Adding UIDs would resolve ambiguity, but they would need to be consistent across rounds/runs/whatever scope is needed for paired evaluation, which requires careful consideration from a privacy point of view (a simple hash of the image/local ID is likely problematic)
* The current mechanism for opting out of transferring the data in question is not participating in the training at all. While a specific opt-out offers more flexibility, it is also an additional opportunity for clients to behave differently (one accidentally mis-configured client is enough) and thus end up with incomplete and non-usable results.

PR #557, opened 2026-09-08
feat: wire per-case prediction return end to end (#526, #527, #528)

Ultimate-Storm:
**Stacked on #534 and #556**, which are green but unreviewed. Opened rather than held, so the work is visible and reviewable as one story; it should merge after its parents.

This is the part that makes the mechanism usable rather than merely present.

## Client

`GT_PredProb_Output_Callback` gains an optional metric writer and, when the site has opted in, returns its **validation** rows through the channel the metrics already use. The writer is attached where the `ClientLogger` is created — outside a swarm run there is none and nothing is returned.

Training rows are deliberately not returned: they're the larger set and nothing in the deliverables needs them.

The whole return is wrapped. Sending predictions is an extra, and a failure there must never cost a training round that has already succeeded.

## Server

`PerCasePredictionCollector` wired into all seven job configs, plus the single-file symlink for `minimal_training_pytorch_cnn`. Inert unless a site opts in — the default.

## Analysis

`compare_regional_finetune` now reads both layouts: `predictions_*.csv` with a `uid` from `predict.py`, and `<site>.csv` with a `row_index` returned through the swarm.

**Keys are namespaced by site.** Every site's `row_index` restarts at 0, so an unnamespaced key would pair a Radboud case with an Utrecht one *and still report it as a paired comparison* — a wrong answer that looks entirely well-formed. There's a test for exactly that.

That change broke an existing assertion. I updated it to the new contract with the reason written into the docstring, rather than reverting the improvement to keep a test quiet.

## Tests

7 new — including cross-site index collision, and pairing a baseline evaluated offline against a fine-tuned model returned through the swarm. Full suite **448 passed, 7 skipped**.


application/jobs/_shared/custom/threedcnn_ptl.py
```diff
@@ -486,10 +486,25 @@ def output_csv(results, epoch: int, csv_filename) -> None:
 
 
 class GT_PredProb_Output_Callback(Callback):
-    def __init__(self, data_module, csv_filename_train, csv_filename_validation):
+    """Write per-case ground truth and class probabilities to local CSV.
+
+    Optionally also returns the *validation* rows to the coordinator, so that
+    D2.5's paired comparison and D3.2's uncertainty ranking can be computed
+    centrally instead of requiring a 722 MB checkpoint to leave the hospital
```
oleschwen: Cannot follow why one would be a replacement for the other.
Transferring individual predictions for the local data allows centrally computing metrics after the training (e.g., metrics that turn out to be useful, but were not implemented at the time of training) based on local (training (not included here?), validation, test (separate issue)) data that remains local. Transferring a checkpoint allows computing metrics using other (e.g., external test) datasets.

PR #558, opened 2026-09-09
docs: add the current decks and an index pointing at the latest

Ultimate-Storm:
Puts the three current pages under `docs/` so they are versioned alongside the code that produced their numbers, and adds `docs/PRESENTATIONS.md` naming which is current.

| File | What it is |
|---|---|
| `presentation_consortium_briefing.html` | 8 slides for the technical meeting |
| `presentation_data_by_site.html` | training data and class distribution, 4 charts |
| `report_wp2_wp3_status_M45.html` | WP2/WP3 status for the committee |

The June experiments deck (`presentation_swarm_results.html`) was already here and is indexed as earlier material.

## New: model comparison

The briefing gains a slide placing the current model against what came before.

| Model | Sites | Malignant AUROC | Macro AUROC |
|---|---:|---:|---:|
| **Swarm · 1DivideAndConquer** | **8** | **0.887** | **0.820** |
| Swarm · 4LME_ABMIL (Jun) | 6 | 0.847 | 0.764 |
| Best single site — UKA · 1DC | 1 | 0.858 | 0.709 |
| Median of 10 single-site models | 1 | 0.658 | 0.601 |
| Weakest single site — RUMC · 1DC | 1 | 0.408 | 0.483 |

**+0.229** malignant AUROC over the median single-site model, **+0.029** over the strongest — which is UKA, the site holding half the consortium's data — and **+0.040** over the previous six-site swarm.

## The caveat is on the slide, not buried

These are three separate evaluations on overlapping but **not identical** challenge subsets — 165 cases for the eight-site run, 300 for the June matrix. The direction is consistent and repeated across them; the exact gaps are not a controlled comparison. Both the slide and the index say so.

### DIFF
--- a/docs/PRESENTATIONS.md
+++ b/docs/PRESENTATIONS.md
@@ -19,7 +19,7 @@ Slide 6 is the model comparison — the eight-site swarm against the previous si
 swarm and against ten single-site models.
 
 ### `presentation_data_by_site.html` — training data by site
-**4 charts · 9 September 2026**
+**5 charts · 9 September 2026**
 
 How much data each hospital brought when it joined, how the consortium total
 accumulated to 34,462 volumes across eight sites, and the class distribution shown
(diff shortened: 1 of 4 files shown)


oleschwen (approved):
result documentation, no MediSwarm or application code affected

PR #559, opened 2026-09-09
539 run preflight check for all odelia models

oleschwen:
This PR 
* extends the weekly/manually triggered run_all_models_preflight_check test by also including ResNet variants
* also verifies that the preflight checks have succeeded (and not just started with the correct model)

Notes
* Passing CI is not sufficient for this PR, as this affects the weekly/manually triggered test. See https://github.com/KatherLab/MediSwarm/actions/runs/34349142014/job/102457820527 for a green run.
* The test is more thorough than before, so the timeout for some of the preflight checks may have been sufficient this time, but may be insufficient in future runs. I'd suggest increasing them only if needed, an unfinished run should be clear from the output.
* Swin3D does not work (https://github.com/KatherLab/MediSwarm/actions/runs/34345519590/job/102446083605), commented this test.
* ResNet variants other than 18 download pre-trained weights on demand (unlike the other models with pre-trained weights contained in the container image). Might be changed in #546.

### DIFF
--- a/scripts/ci/runIntegrationTests.sh
+++ b/scripts/ci/runIntegrationTests.sh
@@ -1048,36 +1048,43 @@ run_3dcnn_training_in_swarm () {
 }
 
 
+_verify_all_model_preflight_check_output () {
+    EXPECTED_OUTPUT_ABOUT_MODEL=$1
+    WHICH_MODEL=$2
+
+    for EXPECTED_OUTPUT in "$EXPECTED_OUTPUT_ABOUT_MODEL" \
+                           "Epoch 0: 100%";
+    do
+        if grep -q --regexp="$EXPECTED_OUTPUT" "$CONSOLE_OUTPUT_FILE"; then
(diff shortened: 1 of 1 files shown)


Ultimate-Storm (approved):
Verified rather than taken on trust:

- Your green run [34349142014](https://github.com/KatherLab/MediSwarm/actions/runs/34349142014) confirms `step 23: Run preflight check for all productive models -> success` on a `workflow_dispatch`. Agreed that PR CI alone could not have shown this.
- The substantive improvement is checking `Epoch 0: 100%`, not just that the right model name appeared. The previous version would have passed on a model that loaded and then died.
- Splitting `_verify_ODELIA_ternary_preflight_check` out is right — that job needs `--model_name` where the challenge jobs derive it.

Flagging Swin3D with a visible `❗` instead of quietly omitting it is the right call, and it is a real finding: 12 of 13 models pass, Swin3D does not. That interacts with #541 — Swin3D is reachable through `MODEL_NAME` for a swarm run, so a site could select a model that cannot train. I'll open an issue for it referencing your line.

Merging.

PR #561, opened 2026-09-09
539 include further tests in weeekly action

oleschwen:
Our integration tests contain several tests that take a while to run and are thus not incluced in the CI for every PR. This PR adds them to the weekly/manual action. Specifically:
* Running two containers in parallel, verifying that names do not clash
* Pushing and pulling images to a local registry (started in a separate container) to verify that the automatic pull as part of starting a client via the startup kit works

Note:
* successful manual run https://github.com/KatherLab/MediSwarm/actions/runs/34470578853/job/102849282721 is needed to pass, green PR is insufficient.
* Running NVFlare unit and integration tests omitted, did not work out of the box in https://github.com/KatherLab/MediSwarm/actions/runs/34360963671/job/102497521378. It's not our code, so we don't need to regularly check if we broke something. If we want to investigate how to get these tests to run in the test infrastructure, we should create a follow-up issue.
* Test that nodes from incompatible startup kits do not connect omitted as it fails due to an unclear error (maybe related to a change in NVFlare version?). Not sure how much value this test still has, postponed to #567
* Consider renaming or splitting action, at least the weekly run on main is not about PR validation, postponed to #570


Ultimate-Storm (commented):
``

oleschwen:
https://github.com/KatherLab/MediSwarm/actions/runs/35601436261/job/106338222668 succeeded.

Ultimate-Storm (approved):
approved

Issue #530, opened 2026-09-03
T3.3: differential privacy Phase 2 — calibrated noise and a privacy budget


Ultimate-Storm:
Grant task T3.3 (M37–M54), UKA-led. Feeds **MS6** (due M54).

Grant text: "implement the technology to optionally include artificial noise in the training process and limit the privacy budget by design."

## Current state — state this honestly in the deliverable

`PercentilePrivacy` is wired into every job's `config_fed_client.conf` (`percentile=10, gamma=0.01`) and is labelled "Apply differential privacy via percentile clipping". **It is not formal differential privacy.** Per `docs/DIFFERENTIAL_PRIVACY.md`, it is a sparsification-and-clipping heuristic (Shokri & Shmatikov, CCS 2015) with no epsilon/delta, no calibrated noise, no budget and no composition tracking.

The gap analysis is already written. Phase 1 is done; this issue is Phase 2.

## Phase 2 checklist — lifted from `docs/DIFFERENTIAL_PRIVACY.md`

- [ ] Implement `CalibratedNoiseFilter` as an NVFlare DXO filter
- [ ] Add an RDP privacy accountant to the training pipelines
- [ ] Add per-round epsilon logging to the metrics CSV
- [ ] Add `PRIVACY_EPSILON_TARGET` to halt training when the budget is exhausted
- [ ] Measure utility impact on the ODELIA benchmark, with and without noise
- [ ] Measure utility impact on STAMP classification

Per-round epsilon logging lands alongside the metric transport work in #525.

## Done when

Utility is reported with and without calibrated noise on the same cohort, and any claim of differential privacy cites a concrete (ε, δ). Until then the deliverable should say "privacy heuristic", not "differential privacy".

Ultimate-Storm:
**Phase 2 is delivered** (#563, merged).

`application/jobs/_shared/custom/privacy_accounting.py` ships a calibrated-noise filter — clip each site's update to a known L2 norm, add Gaussian noise scaled to it — and a self-contained Rényi-DP accountant. Self-contained because Opacus is not in the ODELIA image and adding it would force a rebuild and redistribution to all eight sites.

The privacy budget is now a number the consortium can choose. Over 20 rounds at δ = 10⁻⁵:

| Target ε | Required σ |
|---|---|
| ≤ 10 | 2.4 |
| ≤ 3 | 6.7 |
| ≤ 1 | 18.1 |

Two qualifications carried in the code and in the deliverable text:

[...]

PR #563, opened 2026-09-10
feat: MVPs for active learning (D3.2) and privacy accounting (T3.3 Phase 2)

Ultimate-Storm:
Both TUD-side only — no partner site, no swarm round, no kit rebuild — so both can be exercised now rather than waiting on consortium time.

## D3.2 — acquisition experiment, and a null result worth having

Sweeps a labelling budget over the **165 real per-case predictions** from the eight-site run, comparing entropy and margin sampling against a 200-seed random control.

**Uncertainty sampling did not beat random.** It found *fewer* malignant cases at every budget:

| budget | random | entropy | margin |
|---:|---:|---:|---:|
| 20 | 4.6 ± 1.8 | 2.0 | 3.0 |
| 50 | 11.3 ± 2.5 | 10.0 | 11.0 |
| 100 | 22.2 ± 2.8 | 19.0 | 21.0 |

The calibration diagnostic is the actual finding:

    mean entropy by true class : 0.987 / 0.878 / 0.981   (no-lesion / benign / malignant)
    highest-entropy 30 cases   : 24 no-lesion,  0 benign,  6 malignant
    lowest-entropy  30 cases   : 13 no-lesion, 12 benign,  5 malignant

**The model is most confident on benign cases** — 12 of the only 17 benign cases sit in the 30 *lowest*-entropy cases. Its confidence carries almost no information about the true class, so an uncertainty rule cannot preferentially surface the rare class; it deprioritises it.

That is a property of this model's calibration, not a verdict on active learning — and it is measurable now instead of after ~110 h of consortium time.

**Scope is stated in the module:** this measures acquisition quality only. Whether a model *trained* on those selections is better is untested, and that is the claim D3.2 ultimately needs.

## T3.3 Phase 2 — calibrated noise accounting

RDP accountant for the Gaussian mechanism, per-round epsilon, and PRIVACY_EPSILON_TARGET as an **enforced** budget. Opacus is not in the ODELIA image and adding it would force a rebuild and redistribution to every site, so the accountant is self-contained — for q=1 the Gaussian RDP is closed-form.

Two things stated rather than implied:

- The guarantee is **client-level** DP by output perturbation. It hides whether a *site* participated, not whether a *patient* was in a site's data — that needs DP-SGD (Phase 3). Conflating them is the difference between protecting a hospital and protecting a patient.
- q=1 **over-states** the cost relative to subsampling. Over-stating is the safe direction for a privacy claim.

What it implies for a real run, which is why it was worth building:

    20 rounds, delta 1e-5, client-level
      epsilon <= 10  needs sigma >= 2.4
      epsilon <=  3  needs sigma >= 6.7
      epsilon <=  1  needs sigma >= 18.1

**At 20 rounds a meaningful budget demands noise far above the clipped update norm**, which will not leave a usable model. The honest options are fewer rounds, a weaker stated guarantee, or record-level DP with subsampling amplification — now quantified rather than asserted.

## Tests

23 passing. The privacy tests pin the *direction* of every monotonicity, because an accountant wrong in the optimistic direction licenses a claim that is not true.

Refs #527, #530


Ultimate-Storm:
**Correction: the active-learning null result in this PR is wrong.**

This experiment was run against `workspace/eval_87c5bbee/results/best/`, which holds the 8-site checkpoint evaluated with `--model-name MST`. The checkpoint is **1DivideAndConquer** — its embedded `train_conf` says `ResidualEncoderClsLightning`. The weights loaded partially and produced plausible-looking but meaningless predictions (pooled accuracy 0.533, malignant AUROC 0.476 — chance). That is **E2**.

The corrected re-run already existed in `results/fixed/` (0.752 / 0.887). It was written beside the broken one, and `best` names the *checkpoint kind*, not the better result.

Re-run correctly, **the result inverts — active learning works**:

| Budget | random | entropy | margin |
|---|---|---|---|
| 10 | 2.4 ± 1.3 | **5** | 3 |
| 20 | 4.6 ± 1.8 | **9** | 7 |
| 30 | 6.8 ± 2.1 | **11** | 8 |
| 40 | 8.9 ± 2.3 | **14** | 10 |
| 50 | 11.3 ± 2.5 | **14** | 13 |
| 100 | 22.2 ± 2.8 | 21 | 20 |

Entropy sampling beats random at 5 of 8 budgets and loses at none, roughly doubling malignant yield at the small budgets where a labelling budget actually binds. The advantage disappears at large budgets because the selection covers most of the pool and every method converges.

The calibration story also reverses: the model is **well calibrated** (ECE 0.053, T = 0.954), so the 'confidence is uninformative' reading in this PR's summary is false too.

**The privacy accounting in this PR is unaffected** — it does not read predictions.

Fixed in #568, which also stops this recurring: all three consumers now refuse chance-level predictions naming E2, the bad directory is renamed `best_WRONG_ARCH_MST_do_not_use`, and this script's verdict is now computed from the yields rather than stored as a hardcoded string that could not change when the data did.

PR #568, opened 2026-09-10
fix(d3.2): retract the active-learning null result — it was computed on wrong-architecture predictions

Ultimate-Storm:
**This PR retracts a wrong conclusion I published, and corrects a second one before it shipped.**

## What went wrong

The D3.2 acquisition experiment (#563, **merged**) and the calibration follow-up were both run against `workspace/eval_87c5bbee/results/best/`. That directory holds the 8-site checkpoint evaluated with `--model-name MST`. The checkpoint is **1DivideAndConquer** — its embedded `train_conf` says `ResidualEncoderClsLightning`. The weights loaded partially and produced plausible-looking, meaningless predictions. That is **E2 in our own catalogue**.

The corrected re-run already existed, in `results/fixed/`. It was written *beside* the broken one instead of replacing it, and `best` names the **checkpoint kind** (best vs final), not the better result.

| | `best/` (MST) | `fixed/` (1DC) |
|---|---|---|
| Pooled accuracy | 0.533 | **0.752** |
| Malignant AUROC | 0.476 — chance | **0.887** |
| Lesion-vs-none | 0.375 — below chance | **0.853** |

## Both conclusions reverse

**Active learning works.** Entropy sampling beats random at 5 of 8 budgets and loses at none:

| Budget | random | entropy | margin |
|---|---|---|---|
| 10 | 2.4 ± 1.3 | **5** | 3 |
| 20 | 4.6 ± 1.8 | **9** | 7 |
| 30 | 6.8 ± 2.1 | **11** | 8 |
| 40 | 8.9 ± 2.3 | **14** | 10 |
| 50 | 11.3 ± 2.5 | **14** | 13 |
| 100 | 22.2 ± 2.8 | 21 | 20 |

Roughly double the malignant yield where a labelling budget actually binds. The advantage vanishes at large budgets because the selection covers most of the pool and every method converges — that ceiling is stated rather than hidden.

**The model is well calibrated.** ECE **0.053**, fitted temperature **0.954**, and the apparent reliability inversion is z = 0.12, **p = 0.90 — noise**. E4 as written was false and is withdrawn.

## Why it survived

The headline numbers were never affected — 0.752 / 0.820 / 0.887 come from `analyze_best.py`, which reads `fixed/`. So the deck said 0.887 while the analyses said chance, and nobody reconciled them. **One evaluation with two live answers is the actual defect.**

## Four changes so it cannot recur

1. **All three consumers refuse chance-level predictions** (malignant AUROC < 0.60), naming E2 and pointing at `results/fixed`. Tested, for both row shapes the consumers use.
2. **The superseded directory is renamed** `best_WRONG_ARCH_MST_do_not_use` with a README stating what is wrong with it.
3. **The acquisition verdict is computed, not hardcoded.** It was a fixed string; when the numbers reversed the string did not, so the summary asserted a null result on top of data showing the opposite. *A conclusion that cannot change when the data changes is not a conclusion.*
4. **The calibration verdict no longer conflates** "calibrated beats random" with "recalibration helped". The uncalibrated arm already beats random; recalibration changes selection at 2 of 8 budgets, both times slightly worse.

E4 is rewritten around the real lesson: **when you re-run to correct an evaluation, delete or quarantine the superseded output** rather than leaving it beside the corrected one under a friendlier name.

## Also adds `scripts/evaluation/deck_metrics.py`

Every headline number with bootstrap intervals, so no slide can quote an unreproducible figure:

| | n | AUROC | 95% CI | recall | spec |
|---|---|---|---|---|---|
| ODELIA malignant-vs-rest | 165 | 0.887 | [0.818, 0.945] | 0.486 | 0.977 |
| ODELIA malignant-vs-no-lesion | 148 | 0.903 | [0.837, 0.956] | 0.486 | 0.991 |
| **Duke (independent US)** | 260 | **0.887** | [0.845, 0.926] | 0.662 | 0.976 |

Matched drop Europe → US is **0.016 with heavily overlapping intervals** — consistent with no loss at all on an unseen continent.

Recall of 0.486 at specificity 0.991 is E3: the threshold is placed to avoid false alarms and the model misses cancers it ranks correctly. Fixable by moving the threshold, no retraining.

**24 tests. Full suite 502 passed, 8 skipped.**

#563 needs the same correction — its merged summary still states the null result.


Ultimate-Storm:
**Do not merge — the input data is wrong and the conclusion reverses.**

These predictions came from `workspace/eval_87c5bbee/results/best/`, which holds an evaluation run with `--model-name MST` against a checkpoint that is actually **1DivideAndConquer** (its embedded `train_conf` says `ResidualEncoderClsLightning`). Pooled accuracy there is 0.533. The corrected re-run lives in `results/fixed/` at 0.752 — the numbers the deck and the M45 report already quote.

That is **E2 in our own catalogue**, and I walked into it.

Re-running everything against `results/fixed/` inverts the result:

| | on `best/` (MST, wrong) | on `fixed/` (1DC, correct) |
|---|---|---|
| ECE | 0.180 | **0.053** |
| Fitted temperature | 0.806 | **0.954** |
| Reliability inversion | z=3.48, p=0.0005 | z=0.12, **p=0.90 — within noise** |
| Per-class entropy spread | 0.109 | 0.054 |
| Entropy vs random @30 | 6 vs 6.8 | **11 vs 6.8** |
| Entropy vs random @10 | 1 vs 2.3 | **5 vs 2.3** |

So: the model is **well calibrated**, recalibration is unnecessary, and **uncertainty sampling beats random at 6 of 8 budgets** — roughly double the malignant yield at small budgets. E4 as written is false and will be withdrawn.

Two things I am keeping from this PR, because they earned their place:

1. The significance test on the reliability inversion. On the correct data it returns p=0.90 and **refuses to report the inversion as a finding**. Without it I would have published a second wrong claim.
2. `fit_temperature_checked`, which distinguishes a fitted optimum from a search bound.

I will rewrite this PR against the correct predictions rather than merge it. #563's null result needs the same correction.

PR #572, opened 2026-09-10
fixed output of runIntegrationTests.sh

oleschwen:
This PR fixes confusing output from the data access preflight checks for all models

### DIFF
--- a/scripts/ci/runIntegrationTests.sh
+++ b/scripts/ci/runIntegrationTests.sh
@@ -1091,7 +1091,7 @@ _verify_ODELIA_ternary_preflight_check() {
 
 run_all_models_preflight_check () {
     # requires having built a startup kit and synthetic dataset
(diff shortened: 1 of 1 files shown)


Ultimate-Storm (approved):
Approved.

`run_all_models_preflight_check()` was announcing `[Run] 3DCNN local training...` — a copy-paste label from the neighbouring function. One line, no functional change.

Worth more than it looks. A log line that names the wrong operation is how a green check gets misread, and that has cost real time on this repo lately: `cancelled` rendering identically to `failure` in `gh pr checks` led me to record #550 as "failed 4×" when it was two cancellations and two genuine failures from a stale branch. Anything that makes CI output say what it actually did is worth taking.

CI verified by conclusion rather than by the rendered status: 6/6 SUCCESS, with validate-swarm genuinely running for 18m59s rather than being evicted.

PR #582, opened 2026-09-13
release: 1.8.1 — configure the starting client first (closes #576)

### LINKED ISSUE #576: Swarm starts before the starting client has configured when min_clients < site count (weekly sim failure, F8 in the configure phase)
## What happened

The scheduled weekly run on `main` ([34739314163](https://github.com/KatherLab/MediSwarm/actions/runs/34739314163), 13 Sep 05:02 UTC) failed in **Run 3DCNN training in simulation mode** — a step that had passed on the same code in both PR validations the evening before (#573, #575). Log:

```
05:09:51,376 WARNING - clients ['client_A'] did not configure within timeout but min_clients=1 allows proceeding; they remain as participants and may rejoin in a later round
05:09:51,747 ERROR - Processing error from executor FaultTolerantSwarmClientController: ValueError: invalid model learnable: expect Model type but got <class 'NoneType'>
  File ".../nvflare/app_common/ccwf/client_ctl.py", line 372, in execute
    initial_model = self.shareable_generator.learnable_to_shareable(learnable, fl_ctx)
05:09:51,781 ERROR - Aborting current RUN due to FATAL_SYSTEM_ERROR received: failed to start workflow swarm_controller on client client_A
```

"Did not configure within timeout" arrived **0.4 s** after launch, with `configure_task_timeout = 300`. No timeout elapsed.

## Mechanism

`ServerSideController.control_flow` in the fork (`docker_config/NVFlare`, `app_common/ccwf/server_ctl.py` ~L280–325) waits for the configure task with

```python
required = configure_min_clients if > 0 else min_clients if > 0 else total_clients
self.broadcast_and_wait(task, targets=participating_clients, min_responses=required, ...)
```

With `min_clients = 1` and no `configure_min_clients`, the wait returns as soon as **one** client answers. If that is `client_B`, the server proceeds, logs the misleading "within timeout" warning, and sends the start task to the *starting client* `client_A`, whose configure has not finished — so its persistor has no model and `learnable_to_shareable(None)` raises. Which client answers first is a race between two in-process simulator clients; PR validation won it, the weekly lost it.

This is the F8 mechanism (`docs/SWARM_FAILURE_MODES.md`) inside the configure phase. Consortium runs are protected because `patch_warm_start_job.py` sets `configure_min_clients` to the site count; anything submitted without it is exposed.

## Fixes

1. **Test side (this repo, small PR):** the two 3DCNN simulation scripts add `configure_min_clients = 2` next to the `min_clients = 1` they already inject, so the configure phase is deterministic while runtime tolerance is unchanged.
2. **Controller side (fork, proposed):** after the configure wait, if `self.starting_client` is among the clients without `ready_time`, do not proceed. Re-send the configure task to that client alone and wait up to the remaining `configure_task_timeout`; if it still is not ready, `system_panic` with a reason that names the starting client and points at `configure_min_clients`. A run can tolerate an unconfigured *aggregator*; it cannot tolerate an unconfigured *starting client*, and the current message says the opposite of what happened.

Also worth fixing while there: the warning should say "proceeding at `<required_label>` before all clients answered" rather than "did not configure within timeout" when no timeout elapsed.

## Context

Previous weekly failures (16 Aug – 6 Sep) were in the all-models preflight step and are addressed by #552/#559; this is a different, new failure of the same weekly.

Ultimate-Storm:
Bumps the NVFlare fork submodule to the `MediSwarm-2.8.0` head carrying [NVFlare_MediSwarm#8](https://github.com/KatherLab/NVFlare_MediSwarm/pull/8) (plus the black-formatting follow-up #9), the image version to 1.8.1, and adds the changelog entry.

**Why a point release right after 1.8.0:** with `min_clients` below the number of participating sites the controller could start the swarm on a site that had not finished configuring — an immediate `NoneType` abort or a ten-minute stall and a hung simulator, both seen in CI on 13 Sep (#576). Consortium runs were shielded only because `configure_min_clients` is set to the site count at submission. The starting client is now configured first and on its own; the rest follow with the quorum applied to the whole set.

**Verification plan (recorded here as it completes):**
- [x] PR validation: simulation mode and the two-client swarm ran through the new configure phase — [34782065224](https://github.com/KatherLab/MediSwarm/actions/runs/34782065224) green
- [x] real-kit deploy test on dl3 + dl0 + dl2 with `jefftud/odelia:1.8.1-dev.260913.f7c66bb`: 20/20 rounds, no aborts, PASSED (`mvp1_rc181_baseline_260913_2113`); server log shows the new phase — `sending task swarm_config to starting client TEST_A_1 first`, then `TEST_B_1` under the quorum, `client configuration took 7.2 seconds`; `cross_val_results.json` returned for both sites
- [x] weekly all-models validation on main `a76c691`: [34786412878](https://github.com/KatherLab/MediSwarm/actions/runs/34786412878) green on rerun. The first attempt failed in the minimal-example simulation on a sporadic client stall two rounds after a normal configure phase (#583); the step passed on the rerun and had passed on this code in PR validation.

Closes #576.


Ultimate-Storm:
Post-release evidence: the release-triggered deploy test for v1.8.1 ([34788900263](https://github.com/KatherLab/MediSwarm/actions/runs/34788900263), `run_all`) **trained all six ODELIA models through the new configure phase on real kits on the runners** — MST 210 s, 1DivideAndConquer 540 s, 2BCN_AIM 600 s, 3agaldran 570 s, 4LME_ABMIL 450 s, 5Pimed 480 s, each with both clients registered and server artifacts recovered from the job store. The run is red only because the legacy post-training evaluation step still points at the old Cosmos paths (`UKA_1` on `/mnt/sda1/ODELIA_Challenge_unilateral`), which do not exist on the runners; that is a harness follow-up, not a 1.8.1 problem.

